PR #20, opened 2020-03-06
AppImage

probonopd:
This PR, when merged, will compile this application on [Travis CI](https://travis-ci.org/) upon each `git push`, and upload an [AppImage](http://appimage.org/) to your GitHub Releases page.

Providing an [AppImage](http://appimage.org/) would have, among others, these advantages:
- Applications packaged as an AppImage can run on many distributions (including Ubuntu, Fedora, openSUSE, CentOS, elementaryOS, Linux Mint, and others)
- One app = one file = super simple for users: just download one AppImage file, [make it executable](http://discourse.appimage.org/t/how-to-make-an-appimage-executable/80), and run
- No unpacking or installation necessary
- No root needed
- No system libraries changed
- Works out of the box, no installation of runtimes needed
- Optional desktop integration with `appimaged`
- Optional binary delta updates, e.g., for continuous builds (only download the binary diff) using AppImageUpdate
- Can optionally GPG2-sign your AppImages (inside the file)
- Works on Live ISOs
- Can use the same AppImages when dual-booting multiple distributions
- Can be listed in the [AppImageHub](https://appimage.github.io/) central directory of available AppImages
- Can double as a self-extracting compressed archive with the `--appimage-extract` parameter
- No repositories needed. Suitable/optimized for air-gapped (offline) machines
- Decentralized

[Here is an overview](https://appimage.github.io/apps) of projects that are already distributing upstream-provided, official AppImages.

__PLEASE NOTE:__ For this to work, you need to set up `GITHUB_TOKEN` in Travis CI for this to work; please see https://github.com/probonopd/uploadtool.

If you have questions, AppImage developers are on #AppImage on irc.freenode.net.


probonopd:
An example AppImage for testing is available at https://github.com/probonopd/imagewriter/releases/tag/continuous.

lurch:
Rather than doing everything on TravisCI, I wonder if it might be better to split this into a two-part process, i.e. a PR to add the necessary modifications to the repo to enable an AppImage to be built, and then a separate PR to add TravisCI integration?
Note for example that this project doesn't have anything published to https://github.com/raspberrypi/imagewriter/releases yet :wink: 

P.S. Having the TravisCI download and run arbitrary bash scripts from some 3rd-party repository feels a little bit dangerous? :confused:

probonopd:
Thanks for your review @lurch. I'm happy to make the required changes once I understand them better.

> Rather than doing everything on TravisCI, I wonder if it might be better to split this into a two-part process, i.e. a PR to add the necessary modifications to the repo to enable an AppImage to be built, and then a separate PR to add TravisCI integration?

Would a separate bash script that builds the AppImage (and that gets invoked by `.travis.yml`) fulfill that requirement?

> Note for example that this project doesn't have anything published to https://github.com/raspberrypi/imagewriter/releases yet wink

Well, this PR rectifies that. I have no way to upload the AppImage to the location where the other builds are uploaded, because only Raspberry Pi staff can do that. But it should be easy for them to arrange the AppImage to be uploaded there.

This PR builds an AppImage for each git push, so that changes can be tested immediately ("continuous build").

> P.S. Having the TravisCI download and run arbitrary bash scripts from some 3rd-party repository feels a little bit dangerous?

Where do we "download and run arbitrary bash scripts"? We get the latest Qt from a well-known PPA that has been respected in the industry over years, and we get the AppImage creation tools from the author's download page, since they are distributed in AppImage format. What concrete change would you like to see?

lurch:
> Would a separate bash script that builds the AppImage (and that gets invoked by `.travis.yml`) fulfill that requirement?

Yeah, that's kinda what I was thinking.

> > Note for example that this project doesn't have anything published to https://github.com/raspberrypi/imagewriter/releases yet :wink:

> This PR builds an AppImage for each git push, so that changes can be tested immediately ("continuous build").

I'm not directly involved with the Raspberry Pi Imager tool, so apologies if I overstepped the mark. But based on how other Raspberry Pi projects are developed, I (perhaps naively) assumed that they wouldn't be going for a continuous build / continuous release process.

> > P.S. Having the TravisCI download and run arbitrary bash scripts from some 3rd-party repository feels a little bit dangerous?

> Where do we "download and run arbitrary bash scripts"?

```
  - wget -c https://github.com/probonopd/uploadtool/raw/master/upload.sh
  - bash upload.sh Imager*.AppImage*
```

If your github account gets compromised (either accidentally or maliciously) the `upload.sh` script could be modified to do whatever it likes with our `GITHUB_TOKEN`. Apologies for the tinfoil-hat paranoia, but this doesn't feel like best security practices? :confused:

probonopd:
Do you know a better way that I could use to upload the generated AppImage? (GitLab CI has a feature to store build artifacts, but GitHub?)

lurch:
I can't claim to speak for others, but (IMHO) I still think it would be best if this were split into separate "Create AppImage" and "Add TravisCI integration" PRs (rather than trying to do both jobs in the same PR).

probonopd:
Removed Travis CI. To enable it, simply create a `.travis.yml` file with the following content:

```
language: cpp
compiler: gcc
sudo: require
dist: xenial
  
script:
  - bash -ex linux/appimage.sh

after_success:
  - curl --upload-file Imager*.AppImage https://transfer.sh/Imager-git.$(git rev-parse --short HEAD)-x86_64.AppImage
```

Personally I think this only complicates things (everything was already set up to work end-to-end), but well.

maxnet:
>@maxnet wdyt?

- Not my decision to make, on the question whether AppImage is something that rpi(t) wants to support in the first place.

- From a technical viewpoint I have my doubts AppImage is a foolproof solution to having system utilities run on every Linux distribution.

Keep in mind that Qt 5 is readily available through the normal package system on every recent major Linux distribution.
Even Raspbian has it installed by default (except for a few tiny qml packages), because the pdf viewer uses it as well.
So the advantage of having an AppImage that bundles that is kinda limited.

Do keep in mind that to be able to run this application as a normal user (not root), it needs to be able to talk to the system's UDisks2 daemon over DBus for some operations.
That working properly on every Linux distribution out there (even older versions people still seems to use) would be a bigger point of concern.
And I don't think you can bundle that in an AppImage file....

It also currently requires a recent version of lsblk for disk enumeration.
(Although technically it would be feasible to replace the code with something else.)

probonopd:
Thanks for your quick response.

> Not my decision to make, on the question whether AppImage is something that rpi(t) wants to support in the first place.

From looking at the commit history I was under the impression that you were, but anyhow I assume your opinion has a lot of weight in this project. Who would need to be convinced?

> From a technical viewpoint I have my doubts AppImage is a foolproof solution to having system utilities run on every Linux distribution.

We are not going for "every", but for "most mainstream".

Just as one point of reference, the balenaEtcher tool is widely used and has been distributed for Linux in AppImage format since a very long time. 

> Keep in mind that Qt 5 is readily available through the normal package system on every recent major Linux distribution

Correct, yet some systems (particularly those shipping with Gnome) don't have it installed as part of the default system installation, and not all users like to have to install it (it's many, many megabytes).

Also, I don't think this application is packaged for every distribution out there.

> So the advantage of having an AppImage that bundles that is kinda limited.

Launch a Ubuntu Live ISO, then compare how easy it is to run e.g., balenaEtcher (download one file, make executable, run!) compared to this application (apt-get update as root, download a deb, install as root the full Qt even though only a tiny subset of it is actually needed, install the deb as root,...)

> Do keep in mind that to be able to run this application as a normal user (not root), it needs to be able to talk to the system's UDisks2 daemon over DBus for some operations.

Nothing is preventing the application from doing that. An AppImage is just a self-mounting filesystem; it does not restrict what the application can do. So as long as there is a system's UDisks2 daemon, it should work.

Maybe it would be interesting to see how balenaEtcher, which also ships as an AppImage, solve this aspect?

maxnet:
>From looking at the commit history I was under the impression that you were

Merely a software engineer.
Sure, I can comment on the technical feasibility of things.
But this is a request that can result in a commitment to do more work down the line, then merely accepting this PR.

It could make it appear to end-users that a wide range of Linux distributions is supported, and that they can get a certain level of technical support if things do not work as advertised.
I do not know if that is desired. Sure, it could be argued, that if at least works on some more Linux distribution, that is good to have.
But on the other hand, it could also be said that it would be better if it was encouraged that the community provides their own native packaging for their favorite Linux distribution instead.
That would guarantee that at least the software is tested on that Linux distribution, by the packager, and that there is someone that can provide at least a little bit support if there are any distribution specific bugs.

>So as long as there is a system's UDisks2 daemon, it should work.

The question is whether the UDisks2 daemon version that ships with the distribution is in a usable state.
The functions we call are not so widely used.

What I meant to say, AppImage is pretty much advertised as "just stick every library and everything your application needs in there, no dependency on what the system has"
However reality is that system utilities like this (unlike your average application) do certainly depend on system services.
It is not like we can stick a newer udisks2 in the AppImage, if the one the system provides does not work as advertised.

>Maybe it would be interesting to see how balenaEtcher, which also ships as an AppImage, solve 
>this aspect?

Part of their application runs as root (backend running as root communicates with frontend over a pipe), and they do not support all the functionality we have (e.g. fat32 format).
Different architecture.
Each with different merits.

probonopd:
Yes, everything has pros and cons. Maybe a good middle ground would be to make an AppImage, test it on say the top 5 popular distributions, and clearly state on the download page that it is supported only on those?

ghollingworth:
Serge has already built an AppImage for this, it should be pushed onto the Raspbian server in the next couple of days...

probonopd:
Thanks @ghollingworth - let me know whenever you need any help in anything related to AppImage.

lurch:
Sorry for "wasting your time" @probonopd , I didn't realise @XECDesign had already progressed so far with making an AppImage! :wink:

probonopd:
No worries. Looking forward to the AppImage. Thanks for making this software and its AppImage, very useful.

ghollingworth:
Although it's not on the downloads page yet, you can try it here:

http://downloads.raspberrypi.org/imager/

probonopd:
**Success** with the AppImage on a Fedora 31 Live ISO:

PR #22, opened 2020-03-07
Add i18n feature

YuzukiTsuru:
use a simple header file to achieve i18n feature.
the i18n translates file is in `i18n`

- change in qml file
```
== -> ===
!=  -> !==
```


ghollingworth:
@spl237 Is there a way of providing language choice without having the ugly selection box in the middle of the screen?  Can we get language selection from the installation instead?

spl237:
Depends on the system you are running on; can't say for Windows or Mac, but on Linux you should be able to read the system locale; it's in an environment variable. I'd assume Windows and Mac have similar mechanisms, but don't know what they are.

ghollingworth:
I assume that Qt has a default locale based on the system locale?  So maybe it's just a case of using that?

spl237:
https://doc.qt.io/qt-5/qml-qtqml-locale.html may be relevant....

maxnet:
Cherry picked your .ts

PR #52, opened 2020-04-08
Install icon under share/icons

ghisvail:
Use of `share/pixmaps` is deprecated in favour of `share/icons` with right subfolder depending on the size of the icon (here 128x128). Best would be to provide the icon as an SVG and not worry about sizes depending on the desktop environment.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -224,7 +224,7 @@ else()
     endif()
 
     install(TARGETS rpi-imager DESTINATION bin)
-    install(FILES icons/rpi-imager.png DESTINATION share/pixmaps)
+    install(FILES icons/rpi-imager.png DESTINATION share/icons/hicolor/128x128/apps)
     install(FILES linux/rpi-imager.desktop DESTINATION share/applications)
 endif()
 


lurch:
ping @spl237 and @XECDesign as they know about desktop / packaging things :wink:

XECDesign:
Had to look this one up and this is the clearest explanation I found:

https://lists.fedoraproject.org/pipermail/devel/2011-December/160252.html

Haven't tested the actual change, but LGTM.

spl237:
I'm not 100% sure, but I don't think that will necessarily work. Yes, icons should be in a subdirectory of /share/icons/hicolor/ by default, so they get added to the default icon theme, but I suspect that if a file being used as an icon was installed into share/pixbuf, it will be referenced in the code with an explicit  path to that directory. So I think moving the icon is indeed the right thing to do, but I suspect other changes may be needed elsewhere in the app so that it is still found correctly.

spl237:
You will also need to modify the install process so that the gtk icon cache is updated; otherwise you cannot reference the icon by name in the code. This wouldn't have happened as part of installing files to share/pixbuf, so I suspect that change will also need to be made to the Makefiles to make this happen.

maxnet:
>it will be referenced in the code with an explicit path to that directory.

Not used in application code, only in by the .desktop file.

(In Qt applications it is customary to embed everything the application itself is using inside the executable, instead of having extra files in the file system. Everything listed here: https://github.com/raspberrypi/rpi-imager/blob/qml/qml.qrc ends up in the binary).

spl237:
OK, in that case you probably don't need to do anything other than make sure the gtk icon cache is updated after the file is installed to hicolor - but I suspect you will need to do that to make sure the icon can be found from its name.

XECDesign:
> You will also need to modify the install process so that the gtk icon cache is updated;

That should be automatically handled by dpkg. Again, I haven't tested it, but as far as I know, it monitors icon changes and regenerates cache accordingly at the end.

spl237:
Hmmm - a bunch of installers I have used - including the one I've been working with today - do expiicitly call gtk_update_icon_cache as part of a makefile. I'll try taking that out of the one I've been fiddling with and see if it still works. (Of course, even if dpkg does do it, it does mean that the icon won't be found after a standard build from source - but that's probably not a huge problem...)

ghisvail:
Moving the icon works fine with the Flatpak I am building, btw.

spl237:
Thanks, Serge - you're usually right about this stuff, so I'll defer to you on this one!

XECDesign:
Triggers are something I'm only vaguely familiar with, so I wouldn't have that much faith.

However, just installed the updated package:
```
Reading package lists... Done
Building dependency tree
Reading state information... Done
The following packages will be upgraded:
  rpi-imager
1 upgraded, 0 newly installed, 0 to remove and 2 not upgraded.
Need to get 0 B/266 kB of archives.
After this operation, 3,072 B of additional disk space will be used.
Get:1 file:/home/pi/debs ./ rpi-imager 1.2+nmu1 [266 kB]
Reading changelogs... Done     
(Reading database ... 67775 files and directories currently installed.)
Preparing to unpack .../rpi-imager_1.2+nmu1_armhf.deb ...
Unpacking rpi-imager (1.2+nmu1) over (1.2) ...
Setting up rpi-imager (1.2+nmu1) ...
Processing triggers for mime-support (3.62) ...
Processing triggers for hicolor-icon-theme (0.17-2) ...
Processing triggers for desktop-file-utils (0.23-4) ...
```
So you can see the triggers for for the icon and desktop file both run.

While that was happening, I also ran inotifywait to make sure the `gtk-update-icon-cache` is really getting called, and it is:
```
$ inotifywait `which gtk-update-icon-cache` -m
Setting up watches.
Watches established.

/usr/bin/gtk-update-icon-cache OPEN
```

That doesn't happen when the icon is installed in `pixmaps`, so moving it also fixes that issue.

spl237:
Good work, Serge - thanks for that. In that case, feel free to ignore my ramblings - looks good to merge to me.

PR #98, opened 2020-08-04
update Chinese translation  file rpi-imager_zh_cn.ts

YuzukiTsuru:


lurch:
Is the `.gitignore` added in this PR appropriate @maxnet ? (it seems curiously specific to just two files)
Or would it make sense to delete it from the repo?

maxnet:
Ignoring the .user file (which gets created by Qt creator, and contains local information like the path to the build directory) from `git commit -a` is sensible.

If libarchive creates a .pc in the source directory instead of build directory (haven't verified if it does that. I only run git under Linux, where we use the system's libarchive instead of compiling it ourselves), there may be an issue with the build script, which may be more correct to get fixed upstream.
However from a practical point of view I wouldn't spend too much time in getting issues fixed in other projects that only affect developers rather than end-users. If a one-liner on our side works around the problem, I'm not morally opposed to that. :-)

lurch:
@maxnet Thanks for checking / confirming :+1: 

@YuzukiTsuru Sorry for doubting you, but I've seen a couple of other projects where people have added `.gitignore` files that were specific to their local system rather than the project itself, so I thought I'd double-check.

PR #100, opened 2020-08-05
Add AUR to README

Davoodeh:


tdewey-rpi:
Thanks for the submission, @Davoodeh.

Unfortunately, I'm not minded to merge this change. As we don't maintain the AUR package (nor packages supplied to many other package managers), I don't want to create the impression that we support installations made that way. Arch users installing Raspberry Pi Imager via AUR are using instructions this project hasn't provided, and any issues with that mechanism should be referred to the AUR maintainer.

PR #109, opened 2020-10-04
Allow relative URLs in the repository files

jvprat:
This adds support for using relative URLs in the `icon`, `url` and `subitems_url` fields of the repository files. If the URL in one of these fields is relative, it's preprocessed into an absolute one by using the URL of the containing JSON file as its base. This eases the relocation of repository files without having to change their contents (for example, I could create and test a repository in my local filesystem and then upload them without having to change anything).

### DIFF
--- a/imagewriter.cpp
+++ b/imagewriter.cpp
@@ -301,6 +301,16 @@ QString ImageWriter::fileNameFromUrl(const QUrl &url)
     return url.fileName();
 }
 
+/* Utility function to return an absolute URL from a relative one */
+QString ImageWriter::makeUrlAbsolute(const QUrl &url, const QUrl &baseUrl)
+{
+    if(url.isRelative())
+    {
+        return baseUrl.adjusted(QUrl::RemoveFilename).resolved(url).toString();
+    }
+    return url.toString();
+}
+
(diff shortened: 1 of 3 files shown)


maxnet:
Do would provide backwards compatibility issues if distributions to which we link to using the subitem_urls mechanism start using relative URLs.
rpi-imager does not provide any self-update mechanism, so end-users may still be running older versions that only accept absolute URLs.

sp-jordi-vilalta:
In that case maybe it's a good time to introduce an `oslist_format_version` field at the file root so the imager can warn the user that a newer version should be used

jvprat:
In that case maybe it's a good time to introduce an `oslist_format_version` field at the file root so the imager can warn the user that a newer version should be used

maxnet:
>In that case maybe it's a good time to introduce an oslist_format_version field at the file root so the imager can warn the user 
>that a newer version should be used

But is support for relatively URLs -something that only provides some additional convenience to developers and not to end-users-, something end-users should be forced to upgrade their Imager for?

tdewey-rpi:
Closing this - it's long stale now, with multiple conflicts.

PR #121, opened 2020-11-20
Update Italian language

bovirus:
@maxnet 

Fix previous error (I modified source string and not translation string).
I restore source string and fill-in translated string.

### DIFF
--- a/i18n/rpi-imager_it.ts
+++ b/i18n/rpi-imager_it.ts
@@ -69,8 +69,8 @@
     </message>
     <message>
         <location filename="../downloadthread.cpp" line="277"/>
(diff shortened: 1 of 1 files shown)


lurch:
Also same mistake with `<source>elimina i dati esistenti nell&apos;unità</source>` (maybe others too? :man_shrugging: )

bovirus:
@lurch

Thanks.

I think is not a good idea tod eit directly on github.
It's better to sue QTLinguist to avoid this error.

I will fix in next PR.

maxnet:
>I will fix in next PR.

Just push a new commit to your patch-p1 branch to have it added to this PR?!

PR #124, opened 2020-11-24
Update Italian string for installer

bovirus:
@maxnet 

Update one string for Italian in NSIS installer script.
Please check and merge. Thanks.


maxnet:
Have a new untranslated string in Imager itself as well (update notification)

bovirus:
OK. I will translate also it.
Please merge this change and I will do next one.

PR #134, opened 2020-12-20
Remove usages of old Qt APIs to prepare for Qt6

AaronDewes:
Replace `endl` with `Qt::endl` to prepare for Qt 6 compatibility.


maxnet:
Sorry, cannot accept this until we actually move to Qt 6.

Older versions of Qt 5 -like still shipped with every major Linux distribution- do not have Qt::endl.

PR #140, opened 2020-12-31
UI/Popup: changes the 'NO' button color

marcinciarka:
I switched the colors so the 'NO' button appears to be `secondary` type. I was trying to write an image to my SD card and I pushed the NO button instead of YES.

### DIFF
--- a/MsgPopup.qml
+++ b/MsgPopup.qml
@@ -98,8 +98,8 @@ Popup {
                     msgpopup.close()
                     msgpopup.no()
                 }
-                Material.foreground: "#ffffff"
-                Material.background: "#c51a4a"
+                Material.foreground: "#c51a4a"
+                Material.background: "#ffffff"
                 font.family: roboto.name
                 visible: msgpopup.noButton
                 Accessible.onPressAction: clicked()


MartinMSPedersen:
Good idea and swap Yes/No button places.

maxnet:
>Good idea and swap Yes/No button places.

Note that what should be the order of buttons is something there is much debate over in UI design land.

Whlie Windows platform conventions like the primary action left, Apple human interface guidelines like the primary action right: https://developer.apple.com/design/human-interface-guidelines/macos/windows-and-views/alerts/
Material style / Android like it right: https://material.io/components/dialogs#usage

And what applications do in practice varies a lot.
See also for example: 

https://uxdesign.cc/buttons-placement-and-order-bb1c4abadfcb
https://www.nngroup.com/articles/ok-cancel-or-cancel-ok/
https://uxmovement.com/buttons/why-ok-buttons-in-dialog-boxes-work-best-on-the-right/

tdewey-rpi:
@marcinciarka, @MartinMSPedersen,

Can you take a look at v1.8.1, and see if the changes there have addressed your concerns?

PR #163, opened 2021-03-12
Remove download count retention period from README

mudge:
In order to view the impact of changes such as OS releases over a longer period of time, we have removed the automatic expiry of the anonymous download counts from the rpi-imager stats service.

For reference, the download counts contain only the following data:

    > ZREVRANGE image:2021-03-12 0 1 WITHSCORES
    1) Raspberry Pi OS (32-bit)
    2) 1234
    3) Raspberry Pi OS Lite (32-bit)
    4) 567

Update the README section about telemetry in order to reflect the change.

### DIFF
--- a/README.md
+++ b/README.md
@@ -148,7 +148,7 @@ In order to understand which images and operating systems are most popular and r
 
 This web service is hosted by [Heroku](https://www.heroku.com) and only stores an incrementing counter using a [Redis Sorted Set](https://redis.io/topics/data-types#sorted-sets) for each URL, operating system name and category per day in the `eu-west-1` region and does not associate any personal data with those counts. This allows us to query the number of downloads over time and nothing else.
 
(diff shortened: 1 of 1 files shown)


lurch:
> we've extended the retention period of the anonymous download counts

Are these cumulative download-counts now kept forever then? (which isn't really a "retention period"?) Because without a date-stamp associated with each "count" I guess you wouldn't know when to expire "old counts"?

Just to be clear I'm not personally bothered about any of this statistic-gathering, but I know from experience that people expect things to be very clearly explained.

mudge:
@lurch I've updated the commit message and PR description to hopefully be clearer: we used to explicitly call [`EXPIRE`](https://redis.io/commands/expire) on every day's count so Redis would automatically delete the key after 90 days.

We no longer do this so keys will persist indefinitely by default.

PR #189, opened 2021-03-23
Message strings: SD card -> storage device

andrum99:
The UI has previously been modified to say `storage`, but the message strings still say `SD card` - fix that.

### DIFF
--- a/i18n/rpi-imager_en.ts
+++ b/i18n/rpi-imager_en.ts
@@ -79,7 +79,7 @@
     </message>
     <message>
         <location filename="../downloadthread.cpp" line="740"/>
-        <source>Error reading from storage.&lt;br&gt;SD card may be broken.</source>
+        <source>Error reading from storage.&lt;br&gt;This may be caused by faulty hardware.</source>
         <translation type="unfinished"></translation>
     </message>
     <message>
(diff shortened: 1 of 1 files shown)


maxnet:
Downside of changing strings does is that all translations would need updating as well.
(No longer work if the original string is changed).

andrum99:
> Downside of changing strings does is that all translations would need updating as well.
> (No longer work if the original string is changed).

Apologies - I hadn't realised the translation strings were keyed on the original strings embedded in the code. I've updated the translation files with the new english language versions of the message strings.

(Some of the translations are already out of date even without this change, as not all have had the switch from `SD card` to `storage` applied to the 'choose' button in the main window, and the drive selection window title. So now seems like a good time to update the translations too 😉).

i18n/rpi-imager_de.ts
```diff
@@ -97,7 +97,7 @@ Bitte stellen Sie sicher, dass &apos;Raspberry Pi Imager&apos; Zugriff auf &apos
     </message>
     <message>
         <location filename="../downloadthread.cpp" line="740"/>
-        <source>Error reading from storage.&lt;br&gt;SD card may be broken.</source>
+        <source>Error reading from storage.&lt;br&gt;This may be caused by faulty hardware.</source>
         <translation>Fehler beim Lesen vom Speicher.&lt;br&gt;Die SD-Karte könnte kaputt sein.</translation>
```
lurch, line 101: I think if you'd done this using the QT translation tools (rather than doing it by hand?) it would have automatically marked the relevant translations as "unfinished"?

tdewey-rpi:
Closing as stale, last response was >2 years ago.

PR #265, opened 2021-08-14
Update README.md

andrum99:
Add note that advanced options only works on Raspberry Pi OS, various grammar, spelling and style fixes.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,16 +1,15 @@
 # rpi-imager
 
-Raspberry Pi Imaging Utility
+Raspberry Pi Imager
 
 - Download the latest version for Windows, macOS and Ubuntu from the [Raspberry Pi downloads page](https://www.raspberrypi.org/downloads/).
 - To install on Raspberry Pi OS, use `sudo apt update && sudo apt install rpi-imager`.
 
 ## License
 
-The main code of the Imaging Utility is made available under the terms of the Apache license.
-See license.txt and files in "dependencies" folder for more information about the various open source licenses that apply to the third-party dependencies used such as Qt, libarchive, drivelist, mountutils and libcurl.
+Code is made available under the terms of the Apache license. See license.txt and files in `dependencies` folder for more information about the various open-source licenses that apply to the third-party dependencies used such as Qt, libarchive, drivelist, mountutils and libcurl.
 
-## How to rebuild
+## Build instructions
 
 ### Debian/Ubuntu Linux
 
@@ -45,9 +44,9 @@ cd ..
 sudo apt install ./rpi-imager*.deb
 ```
 
-It should create an icon in the start menu under "Utilities" or "Accessories".
-The imaging utility will normally be run as regular user, and will call udisks2 over DBus to perform privileged operations like opening the disk device for writing.
-If udisks2 is not functional on your Linux distribution, you can alternatively start it as "root" with sudo and similar tools.
+It should create an icon in the start menu under `Utilities` or `Accessories`.
+The imaging utility will normally be run as a regular user, and will call udisks2 over DBus to perform privileged operations like opening the disk device for writing.
+If udisks2 is not functional on your Linux distribution, you can alternatively start it as `root` using sudo and similar tools.
 
 ### Fedora/RHEL/CentOS Linux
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -78,19 +77,17 @@ sudo make install
 
 #### Get dependencies
 
-- Get the Qt online installer from: https://www.qt.io/download-open-source
-During installation, choose a Qt 5.x with Mingw32 32-bit toolchain and CMake.
+- Get the Qt online installer from https://www.qt.io/download-open-source. During installation, choose a Qt 5.x with Mingw32 32-bit toolchain and CMake.
```
lurch, line 80: a) is there a danger that adding a full-stop at the end of the URL means that people might try and include the full-stop in the URL itself?
b) whatever grammar-tweaks you make here, you should probably also duplicate in the Mac OS section?

README.md
```diff
@@ -45,9 +44,9 @@ cd ..
 sudo apt install ./rpi-imager*.deb
 ```
 
-It should create an icon in the start menu under "Utilities" or "Accessories".
-The imaging utility will normally be run as regular user, and will call udisks2 over DBus to perform privileged operations like opening the disk device for writing.
-If udisks2 is not functional on your Linux distribution, you can alternatively start it as "root" with sudo and similar tools.
+It should create an icon in the start menu under `Utilities` or `Accessories`.
+The imaging utility will normally be run as a regular user, and will call udisks2 over DBus to perform privileged operations like opening the disk device for writing.
+If udisks2 is not functional on your Linux distribution, you can alternatively start it as `root` using sudo and similar tools.
```
lurch, line 49: "sudo and similar tools" -> "sudo or similar tools" :man_shrugging:

README.md
```diff
@@ -174,3 +169,5 @@ defaults write org.raspberrypi.Imager.plist telemetry -bool NO
 When using the app, press <kbd>CTRL</kbd> + <kbd>ALT</kbd> + <kbd>X</kbd> to reveal the **Advanced options** dialog.
 
 In here, you can specify several things you would otherwise set in the boot configuration files. For example, you can enable SSH, set the Wi-Fi login, and specify your locale settings for the system image.
+
+**Note:** The advanced options only affect Raspberry Pi OS; support for other operating systems may be added in the future.
```
lurch, line 173: I wonder if this might need to be reworded to clarify that this is related to **flashing** Raspberry Pi OS (regardless of what OS you're currently running RPi Imager on); rather than inadvertently implying that you can only use the "advanced options" if you're running RPi Imager on Raspberry Pi OS itself? :wink: :upside_down_face:

README.md
```diff
@@ -174,3 +169,5 @@ defaults write org.raspberrypi.Imager.plist telemetry -bool NO
 When using the app, press <kbd>CTRL</kbd> + <kbd>ALT</kbd> + <kbd>X</kbd> to reveal the **Advanced options** dialog.
```
lurch, line 169: https://www.raspberrypi.org/blog/raspberry-pi-imager-update-to-v1-6/ says the magic key sequence is ‘Ctrl-Shift-X’ rather than the ‘Ctrl-Alt-X’ written here??

README.md
andrum99: > a) is there a danger that adding a full-stop at the end of the URL means that people might try and include the full-stop in the URL itself?

Yes, but it should be fairly obvious from the fact it is at the end of a sentence that it's only there as punctuation. I would also suggest it is still better to be grammatically correct.

> b) whatever grammar-tweaks you make here, you should probably also duplicate in the Mac OS section?

Good point - I've done that.

README.md
andrum99: I've changed it to `sudo or a similar tool`.

README.md
andrum99: Good point. I've clarified the wording.

README.md
maxnet: >https://www.raspberrypi.org/blog/raspberry-pi-imager-update-to-v1-6/ says the magic key sequence is ‘Ctrl-Shift-X’ rather 
>than the ‘Ctrl-Alt-X’ written here??

Correct.
It's Ctrl-Shift-X

(For some strange reason, those of us who where kids/teens in the 90s may remember ctrl-alt-x better though :-) )

andrum99: Hmm. It's CTRL+SHIFT+X on Windows certainly. `qml/main.qml` says this:

```
Shortcut {
        sequences: ["Shift+Ctrl+X", "Shift+Meta+X"]
        context: Qt.ApplicationShortcut
        onActivated: {
            optionspopup.openPopup()
        }
    }
```

Which I think means SHIFT+CTRL+X **and** SHIFT+<whatever meta is defined as>+X. So I will change the text so it reads CTRL+SHIFT+X.

andrum99: > So I will change the text so it reads CTRL+SHIFT+X.

Done.

maxnet: >Which I think means SHIFT+CTRL+X and SHIFT+whatever meta is defined as+X

Meta is included as alternative key combination because that is "Control" on Mac OS X.

tdewey-rpi:
This is now 2 years stale, and as such I'm going to close it. If the author believes there are further changes to be made against the current HEAD, please submit a new PR and we'll get to it.

Issue #271, opened 2021-08-29
How to change language


kardenwu:
I noticed that there is a folder "`i18n`" in this warehouse. In fact, I am not used to using English, but I am used to my mother tongue, so I would like to ask how to change the language, because I did not find the button

maxnet:
It will use the language of your operating system.

Wowooco:
![image](https://user-images.githubusercontent.com/67304509/131244710-514928ec-222e-4df5-8d9d-0bc7622660d4.png)
What might be the problem? I set it up as you said, but it didn’t take effect!

Wowooco:
> 它将使用您操作系统的语言。
![image](https://user-images.githubusercontent.com/67304509/131244744-bafaf0fa-ba54-4f61-8767-66db8f8720cf.png)
What might be the problem? I set it up as you said, but it didn’t take effect

maxnet:
May need to restart after changing it.

Also note that we only currently have a translation for zh-CN, which I think is mainland China, simplified characters.
If you have some different variant, like zh-TW or zh-HK it will not work.

Wowooco:
> May need to restart after changing it.
> 
> Also note that we only currently have a translation for zh-CN, which I think is mainland China, simplified characters.
> If you have some different variant, like zh-TW or zh-HK it will not work.

I'm sure I set zh_CN, Windows restarted, and other versions of rpi-imager were installed, but it didn't take effect.
System: Windows 10 Pro 21H1
Rpi-imager: 1.6.2

maxnet:
Hmm, seems to be indeed a bug there.

Judging by the output of "powershell" "Get-WinUserLanguageList" it seems that Windows uses "zh-Hans-CN".
(On Linux and MacOS X it does seem to be zh_CN, so if the person that contributed the Chinese translations was using that, he may not have noticed).

Wonder if we should just change rpi-imager_zh_cn.ts to rpi-imager_zh.ts to have it match everything starting with "zh".

Wowooco:
> Hmm, seems to be indeed a bug there.
> 
> Judging by the output of "powershell" "Get-WinUserLanguageList" it seems that Windows uses "zh-Hans-CN".
> (On Linux and MacOS X it does seem to be zh_CN, so if the person that contributed the Chinese translations was using that, he or she may not have noticed).
> 
> Wonder if we should just change rpi-imager_zh_cn.ts to rpi-imager_zh.ts to have it match everything starting with "zh".

[...]

PR #272, opened 2021-08-29
Fix the Chinese language may not work under Windows system

kardenwu:
change rpi-imager_zh_cn.ts to rpi-imager_zh.ts to have it match everything starting with "zh".
Make rpi-imager Chinese can work normally under Windows system😋
issues：[How to change language](https://github.com/raspberrypi/rpi-imager/issues/271)


Wowooco:
According to your request, I amended all "rpi-imager_zh_cn" places to "rpi-imager_zh". But I am very sorry, I tried to install qt, cmake, etc., but I still failed to build this application on my computer.

lurch:
I think `CMakeCache.txt` is an automatically-generated file, and so shouldn't be added to GitHub?

Wowooco:
> I think `CMakeCache.txt` is an automatically-generated file, and so shouldn't be added to GitHub?

Yes, thank you for your reminder. This is the error of the smart prompt, but it is definitely my error

PR #309, opened 2021-11-19
README: Improve custom repository docs

kerel-fs:
- Add description of custom repository file format
-  Add description of official distribution list inclusion procedure
  This text is based on [this comment](https://www.raspberrypi.com/news/raspberry-pi-imager-imaging-utility/#comment-1523756) and just describes the informal process already followed.

PS: Thank you for this awesome tool!


tdewey-rpi:
@kerel-fs Thansk for the submission!

Unfortunately, however, I am not minded to merge this change.

We have been known to change the JSON formats to allow for new features or improved clarity, and I don't want to create an expectation of support of this path. While I doubt anyone will attempt to intentionally "break" custom repository overrides (not least because we use them as part of testing our OS lists), my expectation is that users with a need to use this option will be motivated enough to both find the option and the schema document included in the repository, at which point expanded documentation becomes potentially harmful (especially if we fail to maintain it).

If I see strong demand for custom repository documentation, I may yet change my mind.

PR #331, opened 2022-01-09
Make osdelegate more flexible and robust

mzanetti:
While creating the repository json for the nymea image I noticed that if the json provides an icon that is too large, it can break the delegate. For example:

image

Also I noticed that the description breaks the layout too if it's too long:

image

This patch should keep visuals the same (I think there's a tad more spacing between the lines, keeping Qt's default spacing - if you don't like that I can easily change that). Also it fetches the remote icon now and puts it into 40x40 box like all the icons, scaling if necessariy. A repo json can now feed it it svg or png files of any size and it will always look correct.

Then it splits up the text content in individual entries. For one, this allows to wrap and/or elide as needed and secondly, IMHO this makes the code much easier to work with, without having to construct a huge string with html tags. In order to always fit the content, regardless of how long it is, it will dynamically adjust the delegate height to fit it.
Changed version:
image

If you think it's not good that the delegates height expands to infinity on very long description texts, it can be limited to e.g. 2 lines just by setting `maximumLineCount: 2` on the according Text item.

### DIFF
--- a/main.qml
+++ b/main.qml
@@ -541,97 +541,8 @@ ApplicationWindow {
 
         Item {
             width: window.width-100
-            height: image_download_size ? 100 : 60
+            height: contentLayout.implicitHeight + 24
             Accessible.name: name+".\n"+description
-
-            Rectangle {
-               id: bgrect
-               anchors.fill: parent
-               color: "#f5f5f5"
-               visible: mouseOver && parent.ListView.view.currentIndex !== index
-               property bool mouseOver: false
-            }
-            Rectangle {
-               id: borderrect
-               implicitHeight: 1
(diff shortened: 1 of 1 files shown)


mzanetti:
Ah, the one thing I'm not sure about: The ToolTip. I think this patch should not change anything to the tooltip behavior, however, I couldn't find any entry that seems to produce a tooltip here, so that's untested. Would appreciate if you could point me how to test that one so I can make sure I didn't break it.

mzanetti:
Figured out that tooltips can be provided through the repo json. Tested that too now.

maxnet:
Merged. Thanks.

PR #348, opened 2022-02-04
Add Korean patch

RuffaloLavoisier:
* Add `rpi-imager_ko.ts`
* Update  in `translations.qrc`
```
<file>rpi-imager_ko.qm</file>
```

### DIFF
--- a/i18n/rpi-imager_ko.ts
+++ b/i18n/rpi-imager_ko.ts
@@ -1,0 +1,752 @@
+<?xml version="1.0" encoding="utf-8"?>
+<!DOCTYPE TS>
+<TS version="2.1" language="ko_KR">
+<context>
+    <name>DownloadExtractThread</name>
+    <message>
+        <location filename="../downloadextractthread.cpp" line="167"/>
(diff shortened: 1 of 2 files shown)


RuffaloLavoisier:
If the original is more meaningful than the translation, the original is maintained.

RuffaloLavoisier:
@maxnet 
Hello, Thank you for merge !
but i wonder why did you `close-merge`.

maxnet:
We moved some files around since you opened the PR.
Translations files are now in src/i18n folder instead of i18n.
As a result of that I could not use the "merge" button in github. And cannot edit the commit through the github webinterface either in that case.

Also your PR was missing adding the _ko file to CMakelists.txt

Was more practical to fix things up with a new commit and close-merge there, than figure out how I could fix up things for the new paths.

RuffaloLavoisier:
@maxnet 
I see , thank you for the explanation.

> (Also your PR was missing adding the _ko file to CMakelists.txt)

My mistake..thanks !

Issue #110, opened 2020-10-23
*Suggestion*   [Feature Request] Have an option to choose which raspberry pi's that you have


jack60612:
You would only be shown images for raspberry pi's that you selected
so only images that are compatible with the one that you selected

lurch:
AFAIK, one of the driving factors for RPi-Imager was to make it as easy to use as possible. And it sounds like this suggestion would only make it more complex? :confused: 

[NOOBS](https://github.com/raspberrypi/noobs) and [PINN](https://github.com/procount/pinn) automatically know what Pi they're being run on, so they **do** do the filtering you're asking for.

jack60612:
like an advanced option or something

mobluse:
Some Pis e.g. Pi Zero, Pi 1 A, Pi 4 B can be connected to the computer that runs Imager and thus be automatically identified and also used as an SD card writer/reader.

maxnet:
>Some Pis e.g. Pi Zero, Pi 1 A, Pi 4 B can be 
>connected to the computer that runs Imager and
> thus be automatically identified and also used 
>as an SD card writer/reader.

Correct me if I am wrong.
But I don't think Pi 4 is supported by rpiboot, since on CM4 it requires setting a jumper to disable bootrom and get it to work.

And not sure if the target audience of the models that do support usb boot need help with figuring out which Image is supported by their board.

cillian64:
This is implemented by #645 which will be included in the 1.8.0 release coming shortly

PR #371, opened 2022-02-18
Add support for compat labels

mzanetti:
Currently images list their compatibility in the description which takes away precious space in the description, as well is not unified across different OSes. Looking at the Raspberry PI website, where this became a more important topic with the arm64 images being promoted more popular now, I saw there are nice labels for the compatibility and I thought it may make sense to unify that and came up with this:

image

The attached file is a copy of the upstream os_list json with added compat tags for testing (renamed to .txt because github wouldn't allow attaching .json files).
[os_list_imagingutility_v3.json.txt](https://github.com/raspberrypi/rpi-imager/files/8100743/os_list_imagingutility_v3.json.txt)

 I am not particularly fond of the mechanism to use a comma separated string and splitting that, but QMLs ListModel does not allow injecting arrays just like that. To make it work with `"compatibility": ["A", "B", "C"]` more work and changes would be needed which I didn't want to do without asking if that's something you would even want/accept. If you think the idea in itself is good but would prefer the more proper approach I think I can give that a go when I find some time.

### DIFF
--- a/src/main.qml
+++ b/src/main.qml
@@ -463,6 +463,7 @@ ApplicationWindow {
                     tooltip: ""
                     website: ""
                     init_format: ""
+                    compatibility: ""
                 }
             }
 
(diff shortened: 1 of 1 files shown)


lurch:
I'm not going to comment on whether I think this is a good idea or not (that's not my department :rofl: ), but I did notice that https://www.raspberrypi.com/software/operating-systems/#raspberry-pi-os-32-bit lists "3B" and "3B+" whereas your screenshot shows "3" and "3B+" ?

And I guess this _almost_ relates to #110 as well?

ghollingworth:
I like the idea, but think it would be better to turn it on its head and have a drop-down for compatibility (i.e. for you to choose which device you want to install for and then it only shows compatible images...)

My worry is that in the next 10 years, we'll fill the rest of the space otherwise!!!

tdewey-rpi:
@mzanetti With the introduction of Device Filtering in v1.8.1, I believe this is now largely redundant. If you disagree, please let me know before November 9th, otherwise I'll be minded to close this PR.

PR #373, opened 2022-02-20
Add Enable UART checkbox on Advanced Options Popup

microhobby:
With the Enable UART checkbox selected we will have the property
`enable_uart=1` added to the config.txt.

This is useful if the user already has an usb serial converter
connected and prefers to use the board, or debug it, over serial.

With the 2nd Stage Bootloader UART checkbox select we will have
the property `uart_2ndstage=1` added to the config.txt.

Signed-off-by: Matheus Castello <matheus@castello.eng.br>

### DIFF
--- a/src/OptionsPopup.qml
+++ b/src/OptionsPopup.qml
@@ -117,6 +117,10 @@ Popup {
                         id: chkOverscan
                         text: qsTr("Disable overscan")
                     }
+                    ImCheckBox {
+                        id: chkEnableUart
+                        text: qsTr("Enable UART")
+                    }
                     RowLayout {
                         ImCheckBox {
                             id: chkHostname
@@ -461,6 +465,9 @@ Popup {
         if ('disableOverscan' in settings) {
             chkOverscan.checked = true
         }
+        if ('enableUart' in settings) {
+            chkEnableUart.checked = true
+        }
         if ('hostname' in settings) {
             fieldHostname.text = settings.hostname
             chkHostname.checked = true
@@ -602,6 +609,9 @@ Popup {
 
         if (chkOverscan.checked) {
             addConfig("disable_overscan=1")
+        }
+        if (chkEnableUart.checked) {
+            addConfig("enable_uart=1")
         }
         if (chkHostname.checked && fieldHostname.length) {
             addFirstRun("CURRENT_HOSTNAME=`cat /etc/hostname | tr -d \" \\t\\n\\r\"`")
@@ -788,6 +798,9 @@ Popup {
             var settings = { };
             if (chkOverscan.checked) {
                 settings.disableOverscan = true
+            }
+            if (chkEnableUart.checked) {
+                settings.enableUart = true
             }
             if (chkHostname.checked && fieldHostname.length) {
                 settings.hostname = fieldHostname.text
--- a/src/OptionsPopup.qml
+++ b/src/OptionsPopup.qml
@@ -118,6 +118,10 @@ Popup {
                         text: qsTr("Disable overscan")
                     }
                     ImCheckBox {
+                        id: chk2ndStageUart
+                        text: qsTr("2nd Stage Bootloader UART")
+                    }
+                    ImCheckBox {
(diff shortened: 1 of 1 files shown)


lurch:
IMHO [`uart_2ndstage`](https://www.raspberrypi.com/documentation/computers/config_txt.html#uart_2ndstage) is too special-purpose to be enabled by a general-purpose tool like RPi-Imager.

microhobby:
> IMHO [`uart_2ndstage`](https://www.raspberrypi.com/documentation/computers/config_txt.html#uart_2ndstage) is too special-purpose to be enabled by a general-purpose tool like RPi-Imager.

Well, it's the advanced options, right? 
I like to enable this property for two reasons:

1. To know if my serial connection is working from the power on of the board. (if you have the board connected to a screen you will have some information from the power on, so I have a kind of the same behavior let's say)
2. debug possible problems, or failures, to load anything from the bootloader. (and at this point in my experience `uart_2ndstage` is more effective than when he have the board connected to a screen)

For example if we have some issues and the kernel file is corrupt, whatever the reason, I will have the following in the serial:

image

And nothing, a black screen, on the HDMI output.

I thought about enabling both in the same checkbox, but if you think it's better we can put two checkboxes one for each property. Let me know.

lurch:
If you follow my link to the documentation, it says:
"Be aware that output is likely to interfere with Bluetooth operation unless it is disabled (dtoverlay=disable-bt) or switched to the other UART (dtoverlay=miniuart-bt), and if the UART is accessed simultaneously to output from Linux then data loss can occur leading to corrupted output."

IMHO if you're an advanced enough user to be debugging boot problems, then you're also capable of editing `config.txt` yourself :wink:  (and dealing with the caveats that come with enabling `uart_2ndstage`).

> I thought about enabling both in the same checkbox, but if you think it's better we can put two checkboxes one for each property. Let me know.

That's not my call to make. I'm not directly involved in this project, I'm only an "interested onlooker" (apologies for not making that clear).

microhobby:
> If you follow my link to the documentation, it says: "Be aware that output is likely to interfere with Bluetooth operation unless it is disabled (dtoverlay=disable-bt) or switched to the other UART (dtoverlay=miniuart-bt), and if the UART is accessed simultaneously to output from Linux then data loss can occur leading to corrupted output."

I never had such issues when enabling the `uart_2ndstage`. But if it's in the official documentation, I'll believe it "can occur". This also tell me that make sense of having the two properties in separate checkboxes.

> 
> IMHO if you're an advanced enough user to be debugging boot problems, then you're also capable of editing `config.txt` yourself 😉 (and dealing with the caveats that come with enabling `uart_2ndstage`).
> 

Come on, even advanced users want to make their lives easier. If I'm already using the tool to flash the image on the SD Card, and I have the possibility of having the `config.txt`  automated by the tool itself, IMHO it's much more convenient and easier.

> > I thought about enabling both in the same checkbox, but if you think it's better we can put two checkboxes one for each property. Let me know.
> 
> That's not my call to make. I'm not directly involved in this project, I'm only an "interested onlooker" (apologies for not making that clear).

Ok, so I'll ignore all your comments so far. (just kidding, we know that's not how it works, any feedback from the community is welcome 😅)

lurch:
> Come on, even advanced users want to make their lives easier.

But RPi Imager is also designed to be as easy-to-use as possible (think of e.g. school children), and I think an option labelled "2nd Stage Bootloader UART" (even if it **is** in "Advanced Options") will totally confuse most people (leading to lots of support questions), and do more harm than it does good?

> If I'm already using the tool to flash the image on the SD Card, and I have the possibility of having the config.txt automated by the tool itself, IMHO it's much more convenient and easier.

RPi Imager is only used for flashing the initial image to an SD card - surely by the time you want to mess around with debugging bootloader problems / custom kernels etc. you'll already have booted the OS at least once, by which time you can just do `sudo nano /boot/config.txt` ? :man_shrugging: 

> Ok, so I'll ignore all your comments so far.

:+1:  Makes sense :rofl:

PR #374, opened 2022-02-20
Add option to load SSH keys from GitHub

ShellAddicted:
This PR adds an (advanced) option to automatically load SSH public keys from GitHub
removing the extra step of typing `https://github.com/<username>.keys` in a web browser and copy->pasting the key.

This could be quite useful for the upcoming Network install feature where copy pasting is not even possible.

### DIFF
--- a/src/OptionsPopup.qml
+++ b/src/OptionsPopup.qml
@@ -177,7 +177,7 @@ Popup {
                                     if (chkSetUser.checked && fieldUserName.text == "pi" && fieldUserPassword.text.length == 0) {
                                         chkSetUser.checked = false
                                     }
-                                    fieldPublicKey.forceActiveFocus()
+                                    radioPubKeyString.checked = true
                                 }
                             }
                         }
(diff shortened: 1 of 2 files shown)


NonstickAtom785:
Not sure if commenting is frowned upon but...

Nicely done! I just tested your changes on a local branch, and everything seems to work. Tested the compiled version on WIndows 11, Debian, Endevour OS, and a semi-broken Fedora. They all work correctly and open the browser easily. Might as well add a way to login into the imager with Github.

ShellAddicted:
> Not sure if commenting is frowned upon but...
> 
> Nicely done! I just tested your changes on a local branch, and everything seems to work. Tested the compiled version on WIndows 11, Debian, Endevour OS, and a semi-broken Fedora. They all work correctly and open the browser easily. Might as well add a way to login into the imager with Github.

Thanks.
But what do you mean with "and open the browser easily"?

No browser needs to be opened, that's the point of this PR.

Also, no login is required because as the name implies, public keys are public.

ali1234:
The imager appears to use cloud-init to set up users and keys, so it should be enough to type "gh:username" as your ssh key. In theory this will be stored in the cloud-init config at `/boot/user-data` and cloud-init will fetch the key from github during the first boot process. Have you tried this? I would be quite interested to know if it actually works.

ali1234:
My bad, you have to use a different key in the config, and it relies on ssh-import-id being installed, see: https://cloudinit.readthedocs.io/en/latest/topics/modules.html#ssh-import-id

I still think this would be preferable to fetching the key inside the imaging tool though.

septatrix:
This could generally be expanded to fetch keys from any web URL. Github is not the only platform offering this (and not everyone might use Github). E.g. one could point this to a self-, or company-hosted GitLab or Gitea/ForgeJo which also both support this

PR #402, opened 2022-03-21
Add option to disable user field per-image. Fixes https://github.com/guysoft/pi-imager/issues/8

guysoft:
Hey,
This is a proposal to force in a single-user distribution to grey out the username field and default it to pi.
The json format is:
```json
{
"disable_user":  true
}
```

I am not sure how to add it to the docs here because it looks like an automatically generated HTML, so if you need that an explanation how to do it would be helpful:
https://github.com/raspberrypi/rpi-imager/blob/qml/doc/json-schema/os-list-schema.html

I am open to suggestions how to change this. I already merged it to pi-imager for testing.

This is the current message on octoprint.org, as you can imagine people still do this.:
advanced-password

### DIFF
--- a/src/OptionsPopup.qml
+++ b/src/OptionsPopup.qml
@@ -226,6 +226,7 @@ Popup {
                             TextField {
                                 id: fieldUserName
                                 text: "pi"
+                                enabled: imageWriter.userDisabled()
                                 Layout.minimumWidth: 200
                                 property bool indicateError: false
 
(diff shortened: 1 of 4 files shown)


lurch:
I've got no comment on the contents of this PR, but I'm not sure if we want birthday-greetings in the commit-messages? :confused:

guysoft:
@lurch I can close this commit and open a new one if that is an issue. I cherry-picked it from my fork and didn't notice it took the text with it.

lurch:
> all that is required for it to get merged

I'm not directly involved in this project (which is what I meant with "I've got no comment on the contents of this PR"), so I've got no idea if the change suggested here is actually desired or not; sorry if I gave the wrong impression.

ghollingworth:
This is getting too close to 'fully generalisable options' which is something we're not going to implement.

PR #438, opened 2022-05-11
Enabled mouse select/double click etc

kindid:
From feedback from a user or 2 they wanted mouse selection in text fields as this is common behaviour on their desktop platforms.


tdewey-rpi:
We've addressed this with v1.8.1, and further PRs are always welcome. Thanks for the submission, @kindid

PR #442, opened 2022-05-18
Fix crash on Linux

task-jp:
When one click "CHOOSE OS", app crashes because Qt tries to create
another nam in a thread other than the main thread.
In NetworkAccessManagerFactory::create(), nam->setCache() crashes with
the message below
ASSERT failure in QCoreApplication::sendEvent: "Cannot send events to
objects owned by a different thread. Current thread 0x0x564c398b7550.
Receiver '' (of type 'QNetworkAccessManager') was created in thread
0x0x564c38bcc3f0", file path/to/qt5/qtbase/src/corelib/kernel/
qcoreapplication.cpp, line 553

QNAM::setCache tries to take ownership of the cache which was created in
the main thread.

### DIFF
--- a/src/networkaccessmanagerfactory.cpp
+++ b/src/networkaccessmanagerfactory.cpp
@@ -15,18 +15,20 @@
 /* Configure caching for files downloaded from Internet by QML (e.g. os_list.json and icons) */
 NetworkAccessManagerFactory::NetworkAccessManagerFactory()
 {
-    _c = new QNetworkDiskCache(this);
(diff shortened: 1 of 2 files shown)


maxnet:
With this PR I am getting a different warning on opening OS list:

```
QObject: Cannot create children for a parent that is in a different thread.
(Parent is QNetworkAccessManager(0x55c8a39421e0), parent's thread is QThread(0x55c8a3864fa0), current thread is QQuickPixmapReader(0x55c8a45d0f80)
```

So I think a single QNetworkAccessManager is not the right solution either.

PR #450, opened 2022-06-08
code-quality: fixed all range-loop warnings which were reported from clazy

marcelpetrick:
See: `rpi-imager/src/downloadthread.cpp:1015:14: warning: Missing reference in range-for with non trivial type (QByteArray) [clazy-range-loop]`

### DIFF
--- a/src/dependencies/mountutils/src/linux/functions.cpp
+++ b/src/dependencies/mountutils/src/linux/functions.cpp
@@ -97,7 +97,7 @@ MOUNTUTILS_RESULT unmount_disk(const char *device_path) {
   size_t unmounts = 0;
   MOUNTUTILS_RESULT result_code = MOUNTUTILS_SUCCESS;
 
-  for (std::string mount_dir : mount_dirs) {
+  for (const std::string& mount_dir : mount_dirs) {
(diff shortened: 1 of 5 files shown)


marcelpetrick:
@all Question: are all the currently opened PRs reviewed and processed?  
Not sure if I made something wrong.

maxnet:
>Question: are all the currently opened PRs reviewed and processed?

Eventually.
Making code analyzers happy is not a priority though.

While clazy is technically right in that passing strings by reference instead of by value is theoretically faster, this are non-issues in practice.
Note that copying a string is not an expensive operation with Qt, it only makes copies of a string's contents if its actually modified ( https://doc.qt.io/qt-6/implicit-sharing.html )

marcelpetrick:
@maxnet Thank you for taking time to answer and explain.

I follow a quite different approach: reports from compiler and code-analysis-tools, should be fixed. I agree, for QString, the fix has no positive or negative effect.  
BUT this assumption relies that in the future the implementation details won't change. I can't see into the future.

Also: my professional experience over the last years tell me that migrating a codebase which is almost warning-free is less hazzle than on riddled with. Moving to new targets or toolchains.

I will just leave the PR here and wait for feedback, ok?

Issue #288, opened 2021-10-13
[Feature Request] Set Static IP addresses via advanced options menu


earth4s:
Have an option to be able to set Static IP address in the advanced option menu.

This is a useful feature in headless installs or other embedded situations. Having a pre-assigned IP address takes a lot of the guess work out what the local DHCP assigned the device or if their is not a DHCP implemented on the local network segment.

lurch:
Duplicate of #243 ?

lurch:
Haha, I found it by searching for "static" :wink: 

> Do you want to close my request or keep it separate?

That's up to @maxnet :man_shrugging:

tdewey-rpi:
Closing as a duplicate.

This feature request has been rejected elsewhere because it is unlikely to impact non-technical users, and there's not an immediately obvious way to offer this to non-technical users in a manner that is easily diagnosed and rectified. Use static DHCP leases in a DHCP environment, or a custom-rolled image for more specialised cases.

PR #488, opened 2022-09-24
Advanced settings: Add wired LAN static IP support

### LINKED ISSUE #243: [Feature Request] Add static IPs and file uploading for public key file
Those are the two main features, i wish they were around in rpi-imager.
And maybe a button for advanced settings.

simonkeyd:
This commit enables users to configure a static IP and DNS servers to wired eth0 interface.

This is done by changing popup UI to allow one to specify IP, gateway and DNS servers. And also by adding first run and cloud-init part to configure /etc/dhcpcd.conf accordingly. Finally by ensuring that new settings are also saved.

fixes #243 #288

### DIFF
--- a/src/OptionsPopup.qml
+++ b/src/OptionsPopup.qml
@@ -331,6 +331,68 @@ Popup {
                     }
 
                     ImCheckBox {
+                        id: chkWired
+                        text: qsTr("Configure static wired LAN")
+                        onCheckedChanged: {
+                            if (checked) {
+                                if (!fieldWiredCIDR.length) {
+                                    fieldWiredCIDR.forceActiveFocus()
+                                } else if (!fieldWiredGateway.length) {
+                                    fieldWiredGateway.forceActiveFocus()
+                                } else if (!fieldWiredDNSServers.length) {
+                                    fieldWiredDNSServers.forceActiveFocus()
+                                }
+                            }
+                        }
+                    }
+                    GridLayout {
+                        enabled: chkWired.checked
+                        Layout.leftMargin: 40
+                        columns: 2
+                        columnSpacing: 10
+                        rowSpacing: -5
+
+                        Text {
+                            text: qsTr("IP address/mask (CIDR):")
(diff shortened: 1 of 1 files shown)


simonkeyd:
bump @maxnet @lurch

JinShil:
What about Network Manager?

maxnet:
Not sure if static network configuration is a safe enough feature to include, given the number of less technical users using Imager.
Can disrupt other stuff in network if wrongly configured to use an IP already in use by something else.

Think setting a .local hostname is good enough for most users to reach their Pi, and does not come with risks.

XECDesign:
Any such changes would need to be supported by these scripts first:

https://github.com/RPi-Distro/raspberrypi-sys-mods/blob/master/usr/lib/raspberrypi-sys-mods/imager_custom
https://github.com/RPi-Distro/raspberrypi-sys-mods/blob/master/usr/lib/raspberrypi-sys-mods/init_config

So I think the workflow here should be:

1) Open an issue over at raspberrypi-sys-mods' issue tracker to check whether this feature would be added

1) If so, send a PR there, supporting dhcpcd and NetworkManager (which we plan to move to)

1) If accepted, send a PR here

tdewey-rpi:
We're coming up on the 1 year anniversary since the last comment on this PR.

I'm minded not to accept this change.

I concur with @maxnet that an Avahi advertisement is probably the safest mechanism - the risk of collisions, etc with static IPs is something that requires technical expertise to diagnose and solve - and if a user has those skills, they are probably capable of using either mDNS discovery to identify their Raspberry Pi, or using their router to set a static DHCP lease.

Let me know if you feel I'm incorrect in my assessment before November 9th, otherwise I'll close this PR.

lurch:
There's a bunch of documentation around finding your Pi's IP address at https://www.raspberrypi.com/documentation/computers/remote-access.html#ip-address
And see also https://www.raspberrypi.com/documentation/computers/configuration.html#assign-a-static-ip-address

tdewey-rpi:
Closing, as deadline for comments has long passed.

Issue #350, opened 2022-02-05
Creating am Appimage for rpi-imager


abhas:
I would like to help with creating an updated Appimage for the rpi-image so that it could be used on any GNU/Linux OS with ease and can be recommended as the general purpose image writer for users.

There is already an rpi-image Appimage available via the download site but its very old. I would be glad to help out if the original yml definition of the Appimage is still available somewhere.

Thanks for this lovely utility!

guysoft:
Same here, there isn't any documentation how to build it.

lurch:
There's some (older) info about the old AppImage in #20

maxnet:
I personally think it would be better if those other Linux distributions did some effort themselves and provided native packages.

Also keep in mind that Imager depends on other system services (such as UDisks2) to do its job. And you cannot stick those other services in an AppImage....
So that Imager would magically work on any Linux OS in existence if you would put Imager in a container is a bit too optimistic.

XECDesign:
I don't remember the specifics, but I found that the AppImage build was more trouble than it was worth. Also, it was downloaded only a handful of times.

guysoft:
Hey all, I managed to build an AppImage for my rpi-imager fork. 
You are welcome to test it and tell me if this works for you :)

Due to #200 it wont run on versions of Linux older than Ubuntu 20.04 LTS.

https://github.com/guysoft/pi-imager/releases/tag/v.1.7.1

maxnet:
>Due to https://github.com/raspberrypi/rpi-imager/issues/200 it wont run on versions of Linux older than Ubuntu 20.04 LTS.

Eh, why?

You did not stick a newer Qt version that does have that method in your AppImage?!

guysoft:
Because you need to build rpi-imager on older versions, and it wont build on anything older, see: https://github.com/probonopd/linuxdeployqt#a-note-on-binary-compatibility

maxnet:
>Because you need to build rpi-imager on older versions, and it wont build on anything older

Not with Ubuntu's 5 year old Qt packages, but recall you can get a recent Qt version for Ubuntu 18 from the Qt website.

guysoft:
Possibly. It might work. But at this point I have an AppImage and that's good enough for me.

SharkWipf:
> Also, it was downloaded only a handful of times.

Wouldn't that make sense, given it was, afaik, not listed on the main download page?  
[...]

PR #495, opened 2022-10-15
Add appimagecraft config

TheAssassin:
This PR adds an [appimagecraft](https://github.com/TheAssassin/appimagecraft) build config. appimagecraft is an easy-to-use build script generator that allows building AppImages using short declarative configuration files. It can be used by users to build AppImages with relative ease, but you could also add it to your release cycle.

In my [fork](https://github.com/TheAssassin/rpi-imager) I added a GitHub actions configuration to automatically build AppImages for every commit, creating a so-called [continuous release](https://github.com/TheAssassin/rpi-imager/releases/tag/continuous) using [pyuploadtool](https://github.com/TheAssassin/pyuploadtool). The AppImages are built on Ubuntu 20.04 (which provides a suitable Qt release out of the box), meaning that the AppImage works on most newer distros (tested Ubuntu 22.04 and Fedora 36). In the long term, compatibility may be increased by building the AppImage on an older distro with some custom Qt build following AppImage's [best practices](https://docs.appimage.org/introduction/concepts.html#build-on-old-systems-run-on-newer-systems).
(Side note: the AppImages built in my repository already support [AppImageUpdate](https://github.com/AppImage/AppImageUpdate).)

In case you're interested, I can send a follow-up PR once this one has been merged. I highly encourage you to test the AppImages provided in the continuous release.

References: #350, #20.

screenshot_2022-10-15_12-07-24

P.S.: the appimagecraft config rewrites the desktop file to give the resulting AppImage a sensible file name and make it easily accessible in the user's app launcher (when using some [desktop integration](https://docs.appimage.org/user-guide/run-appimages.html#integrating-appimages-into-the-desktop) tool like, e.g., [AppImageLauncher](https://github.com/TheAssassin/AppImageLauncher)). Can be removed if you don't like it, but it seemed like a good idea to me.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,0 +1,5 @@
+# editor backups
+*.*swp*
+
+# AppImage related
+*.AppImage*
--- a/appimagecraft.yml
+++ b/appimagecraft.yml
@@ -1,0 +1,21 @@
+version: 1
+
+project:
+  name: org.raspberrypi.rpi-imager
+  version_command: git describe --tags
+
+build:
+  cmake:
+    source_dir: src/
+
+scripts:
+  post_build:
+    - |
+      sed -i 's|Name=Imager|Name=Raspberry Pi Imager|' AppDir/usr/share/applications/rpi-imager.desktop
+
+appimage:
+  linuxdeploy:
+    plugins:
+      - qt
+    raw_environment:
+      QML_SOURCES_PATHS: "\"$PROJECT_ROOT\"/src/qmlcomponents/"


TheAssassin:
If you want to test this, just get the branch, download appimagecraft and call it to build an AppImage. See https://github.com/TheAssassin/rpi-imager/blob/qml/.github/workflows/main.yml#L25-L27. Note: it took me a little to figure out which QML modules I needed to install from Ubuntu using trial and error.

tdewey-rpi:
Closing this as stale. There are a few additional items to consider:

1. In this repo we're only including build and packaging infrastructure that we intend to maintain - and I cannot claim to use AppImages on a regular basis.
2. There's nothing to stop you maintaining a decoration-fork if you want to keep packaging this way - I'd note that's likely how Choco, homebrew, etc operate. For the sake of your users, though, I'd encourage you to keep changes to the core program to a minimum, and note that if a user flags that they have installed through this mechanism and we believe the installation mechanism is the root of their issue, I'm likely to close the issue.

TheAssassin:
AppImage is an easy-to-use _upstream_ packaging format. If you don't want to maintain it, I won't either. It makes little to no sense. The maintenance effort is rather low, so I was hoping you'd consider it, maybe as an alternative or replacement to existing packaging.

TheAssassin:
P.S.: This still seems to work fine, I just built a working AppImage. It could easily be integrated into CI. I could add that if you'd merge that then.

tdewey-rpi:
Looks like this PR is pulling a Lazarus.

As I've said in numerous issues and PRs across this project - I'm open to changing my mind if new data comes in. In this case, AppImages have become the front-runner for Imager 2.0, pending some more internal reviews.

TheAssassin:
Please don't hesitate to reach out if you need assistance with this.

tdewey-rpi:
Finally getting around to this - I'm going to make AppImages the primary distribution format for Linux targets with 1.9 in order to include a more recent version of Qt than is shipping in Bookworm.

Raspberry Pi OS will get a convenience deb meta-package that will put the AppImage and launcher files into appropriate places, but otherwise it'll be the Same Thing (tm).

One thing that's not immediately clear to me is how to support aarch64 native-building using AppImageCraft - it looks like I might need to give you a PR for another target?

TheAssassin:
appimagecraft currently does not have a use case and thus lacks support for 64-bit ARM. Some of the tools in the linuxdeploy environment do not support it either due to various reasons. It's on the roadmap.

tdewey-rpi:
Confirmed the changes to linuxdeploy-plugin-qt now package the Wayland client support, and this materially improves UI fluidity on Raspberry Pi 5.

While there is further fettling that could be done to reduce the image size (I'm looking at you, imageformats), the size is comparable to the Windows and macOS distributions, and I consider this an acceptable state for release.

@TheAssassin Would you like a PR against your upstreams?

TheAssassin:
Sure. But I wonder whether you can push to this branch already.

Issue #316, opened 2021-11-30
Adding SSH public keys to "set authorized_keys for pi" in advanced settings doesn't seem to work


raspberrycoulis:
I've been using SSH keys to authenticate with my Raspberry Pi's for a long time now, and until it was supported on the Raspberry Pi Imager I would usually copy a bash file to my SD card that would disable password authentication, turn on public key authentication and then copy my SSH public keys to the `authorized_keys` file before restarting the SSH service within a few seconds.

However, I have now seen that you can paste the SSH public keys to the setting called "set authorized_keys for pi" under the advanced settings menu, but despite pasting the same contents to that box, it does not work when I boot my newly flashed Raspberry Pi.

Does the field in the advanced settings allow for multiple SSH public keys to be added (it is suggested given it's "set authorized_key**s**")? Based on my experience, it does not seem to be working as expected as I cannot SSH into my Raspberry Pi whenever I use this setting, meaning I can only revert back to my previous method which is to manually add my bash script to my SD card, then run it via SSH via the default password, to set it up how I like.

maxnet:
Only supports single line in current version.

geerlingguy:
I had asked for support for _multiple SSH keys_ in #326, but that issue was marked as a dupe of this one, so I'm subscribing to this issue to ask for it as well. I think the title of this issue is a little off, though—if I paste in a file full of pubkeys, the entire content is dumped into the authorized_keys file with a space between each newline. So the first key pasted does work, but others don't, until you manually go in and put all the separate keys on separate newlines again.

So my feature request would be for the public key field in Imager to support newline characters.

pascal456:
This has been closed without comment. I have a question on the changes in the linked commit. Is it correct that multiple keys have to be added in the single-line input field? And that they have to be separated by `\n`?

maxnet:
Don't think there is documentation for that. Had assumed the audience that is using multiple keys, doesn't read docs.

\n = new line

It is assuming you already have a text file with all your public keys, one per line, and are just copying and pasting that to Imager in that single field, in one go.

PR #507, opened 2022-11-25
Make `authorized_keys` field a text area

veleek:
# Change Summary
This change was originally only improving the `authorized_keys` input stuff, but in the process of making it layout properly I ended up doing a bunch of QML cleanup and switching stuff around to what seem to be more standard best practices (as best as I can tell from the QT Quick documentation.

## Make `authorized_keys` a text area instead of a single line field

Currently the `authorized_keys` field is a single line text field in line with a long-ish description.  This makes it difficult to understand what is actually supposed to go in this field in addition to making it very difficult to input multi-line values properly.
I've change this to a TextArea and moved some stuff around and updated descriptions to make this more clear.

*Note: It seems to be very confusing for some what this field actually does.  Even though I've had a raspberry pi setup for years with SSH key auth only, I had no idea what I was supposed to put here or how.  Putting a key in the incorrect format doesn't give any warnings or suggestions.  If I wanted to document this, would it be reasonable to do so inline here, or should that just be linked elsewhere?*

What's the general process for translations.  I made some changes to various fields so the translations would no longer be accurate.

## General QML Cleanup

* Removed rows property on main page, no effect for default LeftToRight layout (only applies for TopToBottom).
* Moved langbar RowLayout to be content of the background rectangle instead of a sibling inside of a parent ColumnLayout (this fixes a runtime error about anchors being used in layout).
* Enabled toggling the lang by on and off instead of just on.
* Removed some manual layout (like calculating offsets to center popup
* Removed redundant font. Family property and set font. Family in root popup.
* Switched popup to ColumnLayout to remove need to do a bunch of anchoring to get stuff layoud out.
* Move popup "title bar" and close button inside of the rectangle that defines the background.
* Tried to use layouts and anchors more consistently in ScrollView to ensure that content fills available space and wraps properly.
* Learned that QML is a pain in the butt
* `Layout.bottomMargin` property for PersistentSettings group box is set to 40 because that's the ONLY way I can get it to display a proper margin INSIDE the scroll view.
* Removed unnecessary text element.

*Original*
image

*New - in general it should be relatively unchanged (other than public keys stuff) but's it's setup better to scale and stuff if necessary.*
image

## Instructions for building without creating a package on Debian

Building and installing a Debian package everytime you make a change is a pain (especially when just working with layout changes).  Updated the Readme to include instructions on building and running on Debian without using the package.

### DIFF
--- a/src/OptionsPopup.qml
+++ b/src/OptionsPopup.qml
@@ -196,7 +196,7 @@ Popup {
                                 enabled: radioPubKeyAuthentication.checked
 
                                 Text {
-                                    text: qsTr("List of authorized SSH public keys.  This will be added to <i>/home/%1/.ssh/authorized_keys</i>.  See the documentation for <i>authorized_keys</i> for details about all supported formats. <br/> <i>Example: ssh-rsa <base64 public key> [optional friendly name]</i>").arg(fieldUserName.text)
+                                    text: qsTr("List of authorized SSH public keys.  This will be added to <i>/home/%1/.ssh/authorized_keys</i>.  See the documentation for <i>authorized_keys</i> for details about all supported formats. <br/> <i>Example: ssh-rsa &lt;base64 public key&gt; [optional friendly name]</i>").arg(fieldUserName.text)
                                     Layout.fillWidth: true
                                     wrapMode: Text.Wrap
                                     color: parent.enabled ? "black" : "grey"
--- a/src/OptionsPopup.qml
+++ b/src/OptionsPopup.qml
@@ -11,13 +11,11 @@ import "qmlcomponents"
 
 Popup {
     id: popup
-    //x: 62
-    x: (parent.width-width)/2
-    y: 10
-    //width: parent.width-125
-    width: popupbody.implicitWidth+60
+    anchors.centerIn: parent
+    width: 600
     height: parent.height-20
     padding: 0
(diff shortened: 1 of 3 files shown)


lurch:
Links to #316 ?

src/main.qml
```diff
@@ -270,75 +269,77 @@ ApplicationWindow {
                     text: qsTr("Keyboard navigation: <tab> navigate to next button <space> press button/select item <arrow up/down> go up/down in lists")
                 }
 
-                RowLayout {
+                Rectangle {
```
veleek, line 272: This section looks MUCH more complicated in the GitHub diff than it actually is.  If you use VSCode (or the Github.dev online one), you'll see that it's just a minor move.

src/OptionsPopup.qml
```diff
@@ -29,414 +27,431 @@ Popup {
     property string cloudinitwrite
     property string cloudinitnetwork
 
-    // background of title
```
veleek, line 32: Ditto for this file, use the Github.dev editor to view the diff and you'll see that the changes are much smaller.

maxnet:
Implemented as textarea in 1.8

veleek:
@maxnet - there were a number of other quality of life fixes included in this change.  Are those included as well or do I need to update this to apply to the head again?

PR #523, opened 2022-12-19
add VSCode tasks file to build and run

veleek:
I dunno if this is something that others would find useful, but this made development work in VSCode a breeze!

### DIFF
--- a/.vscode/tasks.json
+++ b/.vscode/tasks.json
@@ -1,0 +1,37 @@
+{
+    // See https://go.microsoft.com/fwlink/?LinkId=733558
+    // for the documentation about the tasks.json format
+    "version": "2.0.0",
+    "tasks": [
+        {
+            "label": "build and run",
+            "type": "shell",
+            "command": "make; ./rpi-imager",
+            "options": {
+                "cwd": "${workspaceFolder}/build",
+            },
+            "windows": {
+                "options": {
+                    "shell": {
+                        "executable": "pwsh",
+                        "args": [
+                            "-NoProfile"
+                        ]
+                    }
+                }
+            },
+            "problemMatcher": [],
+            "group": {
+                "kind": "build",
+                "isDefault": true,
+            },
+            "runOptions": {
+                "instanceLimit": 1
+            },
+            "presentation": {
+                "showReuseMessage": false,
+                "clear": true
+            },
+        }
+    ]
+}
--- a/.gitignore
+++ b/.gitignore
@@ -1,0 +1,3 @@
+debian
+obj-x86*
+build/


.vscode/tasks.json
```diff
@@ -0,0 +1,37 @@
+{
+    // See https://go.microsoft.com/fwlink/?LinkId=733558
+    // for the documentation about the tasks.json format
+    "version": "2.0.0",
+    "tasks": [
+        {
+            "label": "build and run",
+            "type": "shell",
+            "command": "make; ./rpi-imager",
```
lurch, line 9: Will this still try to run `./rpi-imager` (e.g. a previously-built version) if `make` fails? Might be better to use `make && ./rpi-imager` ?

maxnet:
And "make" only works if you already have a Makefile, e.g. at least have run cmake once.

Don't think this kind of scripting belongs here.
And also think most Qt developers would use Qt Creator instead, as it has better code completion for Qt projects. (e.g. understands Qt specific stuff such as signals and slots)

lurch:
> And "make" only works if you already have a Makefile, e.g. at least have run cmake once.

True, I'd forgotten that this project uses cmake :laughing:

veleek:
Thanks for the feedback guys.  I didn't fill this out completely with dependent scripts to do `cmake` initialization mostly because I wasn't sure if it would even be accepted.  If it's something that would be useful, I'd be happy to do it.

@maxnet - I'm a new Qt developer (but relatively highly experience developer in general) and I honestly had a real rough time getting Qt Creator working particularly well.  The installation process on non-Linux platforms is beastly, and the UI editor was pretty mediocre.  Doing anything that spanned QML and C++ code was just easier in VSCode.  BUT, like I said, I'm not a huge contributor to this project and likely won't be in the future, so take this all with a grain of salt.  Shamless plug for my other PR though #507, which could use a review and has been sitting for a while.

maxnet:
>The installation process on non-Linux platforms is beastly

The same installer installs both Qt creator and Qt SDK.
So would expect that you would also need to run it if you want to use VSCode, or are you using some alternative way?

veleek:
There are a few extensions for VSCode to give syntax highlighting for QML and for most of what I've been doing that was more effective than trying to deal with a visual editor (because I was doing weird adjustments like re-parenting components in different root elements).  For building, I gave up getting Qt Creator working on Windows because there were a bunch of errors that I couldn't work around.  I'm using VSCode via WSL, so for building I didn't need to do anything other than install the QT SDK in my Ubuntu WSL instance.  Qt Creator on Linux via WSL actually works pretty well:

image

But even then, it wasn't rendering any of the UI I was working with properly (the `OptionsPopup.qml`) so I was still better off dealing with the text editor and I'm faster in VSCode because it's my daily driver.

maxnet:
>I'm using VSCode via WSL, so for building I didn't need to do anything other than install the QT SDK in my Ubuntu WSL instance.

Ah, so you are using Ubuntu's Qt with Windows' Linux emulation layer.
Missed that part. :-)

Hard to compile native Windows applications with that.
So unlikely a route most Qt application developers would take.

veleek:
Sorry no, slight misunderstanding.  I ended up using Ubuntu's Qt via WSL because even AFTER getting it installed on Windows natively (a significantly more difficult process, requires an account registration, and not insignificant amount of storage space, 10's of GB or something) I was unable to get it working properly.  It just wouldn't load the project in Qt Creator properly for me.  

The linux version via WSL works correctly, but the rendering of the QML UI doesn't work properly.  See the "Image customization options w/ ComboBox" (GroupBox header) which overlaps the "Set hostname checkbox" (the GroupBox content).  Also the Save button (red button at the bottom) is not the right size and doesn't have any text for some reason.  Additionally everything in that popup is in a ScrollView, but there wasn't an obvious way to scroll the content in the editor to verify that all the margins and padding I was adding were correct.  Since the issue that I was working on was specifically layout related, this basically meant that I didn't get any value out of Qt Creator.  As far as I can tell these are Qt Creator issues, not related to the WSL layer, but it's possible I'm wrong.  Native Linux is the only platform I didn't try.

I even tried installing it natively on MacOS, which was still harder than the linux install and had the same rendering issues (at least some of them).  

However, just to reiterate, I'm not planning on being a major contributor and it's entirely possible that most of the people involved don't have any of these problems.  The core team shouldn't need to bend over backwards to workaround me.  I've got something that works for me (extra extensions in VSCode and my own `tasks.json`) and that's completely fine.  Just wanted to pass on my experience working with the project.  And I did what I could ([see these updates to the README in my other PR](https://github.com/raspberrypi/rpi-imager/pull/507/files#diff-b335630551682c19a781afebcf4d07bf978fb1f8ac04c6bf87428ed5106870f5) to try and smooth the path for anybody else who comes along next!

PR #556, opened 2023-02-21
reduce vulnerabilities

mews-se:


maxnet:
We are only cherrypicking the C++ files from mountutils, drivelist and such, and do not do anything that will cause the dependencies listed in package.json to be imported.

PR #639, opened 2023-10-03
Add lzma dependencies to instructions

tdewey-rpi:
Our CMakeLists will look for lzma, and will fail by default if lzma is not found.

As a response, add lzma development packages to the Fedora and Ubuntu/Debian README sections.

Confirmed this lets CMake configure with default options on WSL2/Ubuntu.


tdewey-rpi:
@maxnet Do you have an opinion on this?

PR #640, opened 2023-10-03
gitignore: Ignore CMake, vscode artefacts.

tdewey-rpi:
While building with CMake and VSCode, I noticed we were not ignoring temporaries and ide-specific configuration.

Add both the default CMake build directory and the vscode private directory to a .gitignore.

This should really be considered a starting point rather than a destination - I fully expect this list to grow as I sound out
the other builds.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,0 +1,2 @@
+build
+.vscode


tdewey-rpi:
@maxnet Separated out, and I reckon it's as fair a starting point as any.

PR #642, opened 2023-10-04
Show options when opening image by file association

### LINKED ISSUE #451: "Advanced options" inaccessible after launching Imager by file association
We are building a custom OS image that is based on [pi-gen](https://github.com/RPi-Distro/pi-gen). During the testing, we found that the "advanced options" button does not show up after launching Imager by double-clicking an image file. It does though when you launch Imager manually and then choose the image file.

cillian64:
If you load a custom image by running `rpi-imager my_image.img.xz` then the Advanced Options button is not shown.  This is triggered if you set a file association to load rpi-imager and then double-click an image file.

The fix is simply to set the visibility on this button on start-up (custom images set on the command-line are loaded before this).

Fixes #451

### DIFF
--- a/src/main.qml
+++ b/src/main.qml
@@ -241,7 +241,7 @@ ApplicationWindow {
                         onClicked: {
                             optionspopup.openPopup()
                         }
-                        visible: false
+                        visible: imageWriter.imageSupportsCustomization()
                         Accessible.description: qsTr("Select this button to access advanced settings")
                         contentItem: Image {
                             source: "icons/ic_cog_red.svg"


tdewey-rpi (approved):
Seems like a solid fix. One thing that's probably worth following up on, however, is the behaviour when you load with a customisable image, but try to select a non-customisable.

cillian64:
> Seems like a solid fix. One thing that's probably worth following up on, however, is the behaviour when you load with a customisable image, but try to select a non-customisable.

I've just checked and this seems to work correctly
- Load a customisable image by file association
- Advanced options button is visible
- Use the choose OS menu to select a non-customisable image
- Advanced options button is no longer visible

PR #652, opened 2023-10-10
i18n: ru: Shorten storage selection label

tdewey-rpi:
Rather than "Choose Memory Device", use "Choose Mem. Device", an acceptable shortening of the phrase which fits inside our controls


maxnet:
@OlesyaGerasimenko any objections?

OlesyaGerasimenko:
Looks good to me.

PR #657, opened 2023-10-13
macOS: Update build instructions to use Homebrew

tdewey-rpi:
Rather than traversing websites and such, instead, let Homebrew manage the install. This simplifies the installation greatly - including managing updates and installation locations.

### DIFF
--- a/README.md
+++ b/README.md
@@ -120,12 +120,13 @@ Note: the CMake integration in Qt Creator is a bit flaky at times. If you made a
     - It is assumed you have an Apple developer subscription, and already have a "Developer ID" code signing certificate for distribution outside the Mac Store. (Privileged apps are not allowed in the Mac store)
 - Install Homebrew (https://brew.sh/)
 - Using Homebrew:
-    - Install qt5: ```brew install qt@5```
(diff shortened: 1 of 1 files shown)


tdewey-rpi:
@cillian64, do you have an opinion on this change?

maxnet:
Note that universal builds you do have to use Qt 6.

cillian64:
I haven't tested it but I think the change is a good idea.

tdewey-rpi:
Ceasing work on this:

1. Homebrew doesn't have a clean solution for pulling x86-64 and arm64 versions of libraries - making it unreliable and non-functional for Universal builds. We could mitigate this by building separately for x86-64 and arm64, or by building _only_ for x86-64 and relying on Rosetta 2, but practically this is a worse experience for all involved
2. Bumping the macOS minimum to 12.0 is required for a Homebrew zlib, which constrains where this tool can be run
3. Even with bumping and pruning arm64, the resultant binary still fails to launch: `qrc:/main.qml:6:1: module "QtQuick" version 2.9 cannot be imported because:
module "QtQml" version 2.9 cannot be imported because:
module "QtQml.WorkerScript" is not installed`

PR #669, opened 2023-10-17
Remove padding from progress bar

cillian64:
Seems like there's a weird Qt bug in the ProgressBar widget where giving it padding means the white progress indicator doesn't actually start at the left edge of the bar or extend to the right edge of the bar.  So just remove the padding from ProgressBar because it's not doing much.

Before: 
image
After: 
image


tdewey-rpi:
Awesome, thanks!

PR #681, opened 2023-10-18
Fix various keyboard focus issues

cillian64:
- Make write confirmation popup modal
- Force focus on the write confirmation popup when it opens
- Force focus on the error popup when it opens
- When the hw select popup opens, force focus on the ListView instead of on the currentItem, because the latter doesn't work when the list is empty.
- On embedded, when the settings popup closes, force focus back to the usecustomisation dialog


tdewey-rpi (approved):

tdewey-rpi:
Awesome!

PR #682, opened 2023-10-18
i18n: Remove broken translations

tdewey-rpi:
As part of changing the reference string, these translations will require a rework.

Without this change, any translations would use both the wrong string _and_ fail to get relfected in the UI.


cillian64 (approved):
lgtm

PR #692, opened 2023-10-20
Added missing dependencies for Ubuntu 22.04.3 LTS

randunel:
Fixes compiling on a fresh installation of Ubuntu 22.04.3 LTS.

### DIFF
--- a/README.md
+++ b/README.md
@@ -20,8 +20,8 @@ For the embedded (netboot) build see also "embedded/legal-info" for more informa
 Install the build dependencies:
 
 ```
-sudo apt install --no-install-recommends build-essential devscripts debhelper cmake git libarchive-dev libcurl4-gnutls-dev liblzma-dev \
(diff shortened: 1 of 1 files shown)


maxnet:
The build system should be patched instead to not make drm a hard dependency.
DRM is only needed for embedded edition of Imager. (can probably consider systems without Qt widgets to be embedded in CMakelists.txt)

randunel:
Agreed, libdrm as a dependency seemed a bit odd. I'll leave it up to the package maintainers to decide how to proceed.

maxnet:
>Agreed, libdrm as a dependency seemed a bit odd.

Yeah, on embedded it is a bit hackery as well.
Reason is that vc4-kms creates multiple DRI devices /dev/dri0 /dev/dri1 and we need to test which one actually can output stuff to screen, before starting Qt.
Not something that is necessary on any other device than a Pi.

PR #694, opened 2023-10-20
qml: Recommend the 'top' OS

tdewey-rpi:
We're going to take advantage of the convention of always putting the recommended OS first in the list, and append "(Recommended)" to it's description.

This moves the recommendation out of the JSON, and makes the recommendation durable to filtering.

### DIFF
--- a/src/main.qml
+++ b/src/main.qml
@@ -1654,6 +1654,19 @@ ApplicationWindow {
             if (oslist === false)
                 return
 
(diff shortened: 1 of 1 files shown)


tdewey-rpi:
Merging this - local testing reckons it's happy enough. Before next rpi-release, we'll need to update the JSON file to remove the `recommended` wording in the descriptions, and let rpi-imager take over.

PR #695, opened 2023-10-20
icons: Update macOS, Linux

tdewey-rpi:
Provide updated macOS and Linux icon files, as supplied by the comms team.

Windows forms are to follow.


tdewey-rpi:
@maxnet I _believe_ these are in the right format - but we don't yet have a Windows icon to match.

maxnet:
Are you sure you want Linux to have an Apple style icon?

Is the plan to replace the icons of all other apps in RPI OS as well?
If not, this may look out of place...

tdewey-rpi:
Tagging @jackbenwillis for visibility.

We had an internal discussion, and the desire was to retain a single icon across platforms that also complied with the Apple HI guidance.

maxnet:
>Windows forms are to follow.

Does icns2ico fails to run, or what is the reason Windows will come later if it is going to be same icon?

tdewey-rpi:
> > Windows forms are to follow.
> 
> Does icns2ico fails to run, or what is the reason Windows will come later if it is going to be same icon?

Windows uses different default icon sizes - while macOS wants NPOT dimension alternatives, Windows wants NPOT plus a few intermediates. I've pinged @jackbenwillis to see if we can get those created and verified - rather than me simply taking 64x64 and scaling down in raster.

cillian64 (approved):

tdewey-rpi:
Windows icons updated. Thanks @jackbenwillis.

On Windows, we previously supplied 5 resolutions. Using the assets from Jack, I've expanded that to 12 resolutions - which should materially improve the look on HiDPI and high resolution displays.

PR #703, opened 2023-10-23
QML: Move 'Recommended' into Application logic.

tdewey-rpi:
We're going to take advantage of the
convention of always putting the recommended OS first in the list, and
append "(Recommended)" to it's description.

This moves the recommendation out of the JSON, and makes the
recommendation durable to filtering.

### DIFF
--- a/src/config.h
+++ b/src/config.h
@@ -8,10 +8,10 @@
 
 
 /* Repository URL */
-#define OSLIST_URL                        "https://downloads.raspberrypi.org/os_list_imagingutility_v3.json"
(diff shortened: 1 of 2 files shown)


cillian64:
This seems to overlap with #694 (needs a rebase?). Not sure why github is saying it merges cleanly

cillian64 (approved):
Anyway, looks good (once the v4 manifest is uploaded)

tdewey-rpi:
Manifest V4 is added and populated. Tested this change in WSL, and it appears to work a treat. V4 also points to v1.8.1 as an update, but once we release v1.8.2 we'll point it at that.

PR #704, opened 2023-10-23
config: Use ImageList v4

tdewey-rpi:
In order to ensure the 'Recommended' tag shows up for older versions of the Imager, we should migrate to using V4 for the version where we move the tag out of the JSON file.

### DIFF
--- a/src/config.h
+++ b/src/config.h
@@ -8,10 +8,10 @@
 
 
 /* Repository URL */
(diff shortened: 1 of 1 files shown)


cillian64 (approved):
lgtm once we have the v4 JSON manifest

tdewey-rpi:
Manifest V4 is added and populated. Tested this change in WSL, and it appears to work a treat. V4 also points to v1.8.1 as an update, but once we release v1.8.2 we'll point it at that.

PR #705, opened 2023-10-25
Update german translation

rejas:
... after seeing 1.8.1 in action I thought of making some translation clearer


tdewey-rpi:
Danke, @rejas. This will go into 1.8.2.

PR #709, opened 2023-10-27
Also customize image if only a cloudinit payload exists

RobinGeuze:
Currently you can configure cloudinit-userdata and cloudinit-networkconfig via the CLI, however due to a check around the customizeImage call it would never be applied.


maxnet (approved):

tdewey-rpi:
Thanks for the PR, @RobinGeuze!

PR #712, opened 2023-11-01
Update Ukrainian translation

Vovkiv:
Updated strings where "image" were replaced with "os", re-translated some strings and fixed typos.


tdewey-rpi:
Dyakuyu, @Vovkiv. Verified manually.

PR #713, opened 2023-11-02
qml: Disable disk cache entirely

### LINKED ISSUE #707: imager ignore the setting i put in (v1.8.1)
i use the advance setting, and its seems to ignore all the setting i put in
like hostname and user and password, it run as Admin on win 10

tdewey-rpi:
This is a compromised change. Essentially, I'm working around a poorly-understood failure mode in Qt5, where the disk cache can occasionally fail to invalidate, resulting in stale compiled units being used.

This will increase application start times in favour of correctness.

Likely to resolve #707

### DIFF
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -140,6 +140,16 @@ int main(int argc, char *argv[])
     }
 
     QCoreApplication::setAttribute(Qt::AA_EnableHighDpiScaling);
+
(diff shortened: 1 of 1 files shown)


cillian64:
Just tested on a Pi 4 and the difference in start time from this patch is negligible.

cillian64 (approved):

PR #716, opened 2023-11-06
qml: Return empty object on JSON failures

### LINKED ISSUE #714: [BUG] Unable to write custom images / format as FAT32 without Internet
**Describe the bug**

It is no longer possible to write custom images or format as FAT32 without Internet.
As also noticed by end-users: https://forums.raspberrypi.com/viewtopic.php?t=359002

**To Reproduce**

- Disconnect cable or tear down network interface.
- Start Imager
- Press "choose OS"
- Empty screen there

**Additional context**

Imager was originally designed to still offer the option to write a custom .img file, and format as FAT32, even if the user doesn't have Internet and downloading the os_list therefore fails.
This was done by hard coding those options in the osmodel ListModel in main.qml and pre-pending additional items to it on download.

Commit https://github.com/raspberrypi/rpi-imager/commit/3f665e01b40dbf0b9496bb853d6250e7e80af877#diff-6b6066b8c2cfa063e0014db476a9967f054fc913bb71de97644e91e91e3660d3R1400 changes this to only append the custom image options after the OS list is downloaded.
This renders Imager unusable to users that cannot directly use the RPI server for any reason. (Could be no Internet, but also that they are running an old operating system that does not have the right SSL certificates, or that they cannot reach the server directly to firewall problems, which they may have no control over.)

If this change in behavior was intended then at least the commit log should mention that.
Failing that, I consider this a bug.

tdewey-rpi:
When we changed the mechanism for appending the special flash options to the OS list, this made those items dependent on a JSON parsing success.

Unfortunately, there's a number of ways that JSON parsing can fail. And when it did, it took out the special items with it.

So, let's make failure return an empty object - which is safe with all current callers - and let the subsequent iterators catch the empty case.

Tested on WSL, in a no-internet scenario.

Resolves #714

### DIFF
--- a/src/main.qml
+++ b/src/main.qml
@@ -1469,27 +1469,37 @@ ApplicationWindow {
         if (!oslist) {
             if (!"os_list" in o) {
                 onError(qsTr("Error parsing os_list.json"))
-                return false
+            } else {
+                oslist = o["os_list"]
+
+                if (hwTags != "") {
+                    filterItems(oslist, JSON.parse(hwTags), hwTagMatchingType)
+                }
+                checkForRandom(oslist)
+
+                /* Flatten subitems to subitems_json */
+                for (var i in oslist) {
+                    var entry = oslist[i];
+                    if ("subitems" in entry) {
+                        entry["subitems_json"] = JSON.stringify(entry["subitems"])
+                        delete entry["subitems"]
+                    }
+                }
+
+                return oslist
             }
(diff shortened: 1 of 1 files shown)


tdewey-rpi:
Not safe to merge yet - testing was flawed. I asked software to isolate the laptop, but that's produced a different results to physically disconnecting cables.

PR #717, opened 2023-11-06
Remove default asignees from issue templates

cillian64:
We'll set assignees when actually working on issues instead of having every issue assigned when created.


tdewey-rpi (approved):
Approved correcting my overzealous template.

PR #719, opened 2023-11-09
qml: move OS list fetch to backend

### LINKED ISSUE #714: [BUG] Unable to write custom images / format as FAT32 without Internet

tdewey-rpi:
- Simple implementation of OS list fetching in backend
- Replace frontend OS list fetching by calls to backend
- OS list updates are brought in asynchronously, avoiding excessive UI blockage.
- "Erase" and "Custom" OS list options are always present, even in a no-internet scenario

Based-On: cillian64/rpi-imager/oslist_backend
Resolves #714

### DIFF
--- a/src/imagewriter.cpp
+++ b/src/imagewriter.cpp
@@ -472,6 +472,12 @@ namespace {
 } // namespace anonymous
 
 
+void ImageWriter::setHWFilterList(const QByteArray &json) {
+    QJsonDocument json_document = QJsonDocument::fromJson(json);
+
+    _deviceFilter = json_document.array();
+}
+
 void ImageWriter::handleNetworkRequestFinished(QNetworkReply *data) {
     // Defer deletion
     data->deleteLater();
--- a/src/imagewriter.cpp
+++ b/src/imagewriter.cpp
@@ -486,18 +486,18 @@ void ImageWriter::handleNetworkRequestFinished(QNetworkReply *data) {
         auto httpStatusCode = data->attribute(QNetworkRequest::HttpStatusCodeAttribute).toInt();
 
         if (httpStatusCode >= 200 && httpStatusCode < 300) {
-            auto responseDoc = QJsonDocument::fromJson(data->readAll()).object();
+            auto response_object = QJsonDocument::fromJson(data->readAll()).object();
 
-            if (responseDoc.contains("os_list")) {
+            if (response_object.contains("os_list")) {
                 // Step 1: Insert the items into the canonical JSON document.
                 //         It doesn't matter that these may still contain subitems_url items
                 //         As these will be fixed up as the subitems_url instances are blinked in
                 if (_completeOsList.isEmpty()) {
                     std::lock_guard<std::mutex> lock(_deviceListMutationMutex);
-                    _completeOsList = QJsonDocument(responseDoc);
+                    _completeOsList = QJsonDocument(response_object);
                 } else {
                     std::lock_guard<std::mutex> lock(_deviceListMutationMutex);
-                    auto new_list = findAndInsertJsonResult(_completeOsList["os_list"].toArray(), responseDoc["os_list"].toArray(), data->url());
+                    auto new_list = findAndInsertJsonResult(_completeOsList["os_list"].toArray(), response_object["os_list"].toArray(), data->url());
                     auto imager_meta = _completeOsList["imager"].toObject();
                     _completeOsList = QJsonDocument(QJsonObject({
                         {"imager", imager_meta},
@@ -505,7 +505,7 @@ void ImageWriter::handleNetworkRequestFinished(QNetworkReply *data) {
                     }));
                 }
 
-                findAndQueueUnresolvedSubitemsJson(responseDoc["os_list"].toArray(), _networkManager.get());
+                findAndQueueUnresolvedSubitemsJson(response_object["os_list"].toArray(), _networkManager.get());
                 emit osListPrepared();
             } else {
                 qDebug() << "Incorrectly formatted OS list at: " << data->url();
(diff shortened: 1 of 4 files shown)


cillian64 (approved):

maxnet:
My first thoughts when seeing this code:

- note that we currently use caching for downloads by the qml code. (configured in the networkaccessmanagerfactory)
If the same json file is fetched twice (e.g. because of model change), it is not downloaded twice.
You are creating your own networkaccessmanager objects but I do not see you emabling caching.
Is that intended?

- don't understand the loose functions in anonomous namespace concept. Would be tempted to make them protected class member functions instead, and let them access variables they need like _networkManager directly, instead of passing through function argument.

- not sure if the mutex stuff is strictly necessary.
Imagewriter should be running in same thread as qml/UI.
Stuff we have running in other threads signals main thread through Qt's signal/slot mechanism that posts messages that are read later in main thread's event loop (qt's [queued connection](https://doc.qt.io/qt-6/threads-qobject.html#signals-and-slots-across-threads) method)

tdewey-rpi:
> My first thoughts when seeing this code:
> 
>     * note that we currently use caching for downloads by the qml code. (configured in the networkaccessmanagerfactory)
>       If the same json file is fetched twice (e.g. because of model change), it is not downloaded twice.
>       You are creating your own networkaccessmanager objects but I do not see you emabling caching.
>       Is that intended?

Yes - the intent of this PR is that we're only fetching the various JSON files once per load. I can't see that caching will materially improve that, and would expect JSON servers to perform their own edge caching as is appropriate for their case.

> 
>     * don't understand the loose functions in anonomous namespace concept. Would be tempted to make them protected class member functions instead, and let them access variables they need like _networkManager directly, instead of passing through function argument.

Coding style. I prefer helper functions to _not_ be part of the class, and constrained entirely to the compile unit they're present in.

> 
>     * not sure if the mutex stuff is strictly necessary.
>       Imagewriter should be running in same thread as qml/UI.
>       Stuff we have running in other threads signals main thread through Qt's signal/slot mechanism that posts messages that are read later in main thread's event loop (qt's [queued connection](https://doc.qt.io/qt-6/threads-qobject.html#signals-and-slots-across-threads) method)

When I open rpi-imager in GDB, I'm seeing multiple Qt threads in play, and with the poor documentation around QJson's mutability I decided to not take any chances. Additionally, the QNetworkAccessManager documentation suggests there is parallel activity in play, but doesn't specify the threading model of it's finished() callback:

image

tdewey-rpi:
Latest version of this PR resolves a bug around resolving redirected resullts, and moves the HW filter completely into the backend too. Tried it out in WSL, and it certainly _felt_ snappier, but I've not done any measurement.

src/imagewriter.cpp
```diff
@@ -416,6 +424,212 @@ void ImageWriter::setCustomOsListUrl(const QUrl &url)
...
+                    }
+                } else {
+                    // No devices tags, so work out if we're exclusive or inclusive filtering!
+                    if (inclusive) {
+                        returnArray.append(ositem);
+                    }
+                }
+            }
+
+            //returnArray += ositemObject;
```
cillian64, line 583: Was this line meant to be deleted?

maxnet:
>but doesn't specify the threading model of it's finished() callback

Think you haven't fully discovered the beauty of Qt's signal/slot mechanism yet. Free thread safety. ;-)

It knows when the slot receiver is in a different thread then the signal emitter. (QObjects store thread affinity: https://doc.qt.io/qt-6/qobject.html#thread )
If it is, a call to "emit finished()", will not be processed as a direct callback.
Instead it will post a message to the message loop of the thread of the receiver.
Qt's queued connecting thing.
So your finished() slot will always run in your own thread.

tdewey-rpi:
> > but doesn't specify the threading model of it's finished() callback
> 
> Think you haven't fully discovered the beauty of Qt's signal/slot mechanism yet. Free thread safety. ;-)
> 
> It knows when the slot receiver is in a different thread then the signal emitter. (QObjects store thread affinity: https://doc.qt.io/qt-6/qobject.html#thread ) If it is, a call to "emit finished()", will not be processed as a direct callback. Instead it will post a message to the message loop of the thread of the receiver. Qt's queued connecting thing. So your finished() slot will always run in your own thread.

That's actually a good point! I'll go ahead and drop the Mutexes.

src/imagewriter.cpp
tdewey-rpi: Yes, it was.

PR #729, opened 2023-11-17
i18n: Japanese minor wording corrections

kznrluk:
The original translation is correct, but this can be made better.

For your reference, I am attaching the number of results from the search with the original wording and my wording. It is one reference.

Screenshot 2023-11-17 125247
Screenshot 2023-11-17 125304

---
Screenshot 2023-11-17 124945
Screenshot 2023-11-17 125017


tdewey-rpi:
Thank you for your submission, @kznrluk.

While I generally avoid stylistic change PRs, I handle translations for languages in which I'm not a native-equivalent speaker differently, and as a result I am happy to accept your patch.

Arigato gozaimashita.

PR #733, opened 2023-11-18
Fix file being passed to --repo

AaronDewes:
Since #719, you can no longer use a local json file with --repo (At least in Qt 6). A network request to a file:// url has statusCode 0, which caused this check to fail previously.

### DIFF
--- a/src/imagewriter.cpp
+++ b/src/imagewriter.cpp
@@ -488,7 +488,7 @@ void ImageWriter::handleNetworkRequestFinished(QNetworkReply *data) {
     if (data->error() == QNetworkReply::NoError) {
         auto httpStatusCode = data->attribute(QNetworkRequest::HttpStatusCodeAttribute).toInt();
 
-        if (httpStatusCode >= 200 && httpStatusCode < 300) {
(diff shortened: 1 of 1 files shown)


tdewey-rpi:
D'oh. I just noticed this and made a duplicate patch. Thanks for the PR, @AaronDewes!

src/imagewriter.cpp
```diff
@@ -488,7 +488,7 @@ void ImageWriter::handleNetworkRequestFinished(QNetworkReply *data) {
     if (data->error() == QNetworkReply::NoError) {
         auto httpStatusCode = data->attribute(QNetworkRequest::HttpStatusCodeAttribute).toInt();
 
-        if (httpStatusCode >= 200 && httpStatusCode < 300) {
+        if (httpStatusCode >= 200 && httpStatusCode < 300 || httpStatusCode == 0) {
```
lurch, line 491: I assume this works as-is, but would it be better to explicitly group the logical operators? i.e.
```
if ((httpStatusCode >= 200 && httpStatusCode < 300) || httpStatusCode == 0) {
```

PR #737, opened 2023-11-20
Show sizes less than 1GB in more detail

AaronDewes:
PR is quite large due to the autogenerated .ts updates, the actual change is in `main.qml`.

grafik
grafik

### DIFF
--- a/src/i18n/rpi-imager_zh-TW.ts
+++ b/src/i18n/rpi-imager_zh-TW.ts
@@ -213,27 +213,47 @@
 <context>
     <name>ImageWriter</name>
     <message>
-        <location filename="../imagewriter.cpp" line="248"/>
+        <location filename="../imagewriter.cpp" line="260"/>
         <source>Storage capacity is not large enough.&lt;br&gt;Needs to be at least %1 GB.</source>
         <translation>儲存容量不足。&lt;br&gt;至少需要 %1 GB。</translation>
     </message>
     <message>
-        <location filename="../imagewriter.cpp" line="254"/>
+        <location filename="../imagewriter.cpp" line="266"/>
         <source>Input file is not a valid disk image.&lt;br&gt;File size %1 bytes is not a multiple of 512 bytes.</source>
         <translation>輸入檔案不是有效的磁碟映像檔。&lt;br&gt;檔案大小 %1 位元組不是 512 位元組的倍數。</translation>
     </message>
     <message>
-        <location filename="../imagewriter.cpp" line="442"/>
+        <location filename="../imagewriter.cpp" line="603"/>
+        <source>Erase</source>
+        <translation type="unfinished">清除</translation>
+    </message>
+    <message>
+        <location filename="../imagewriter.cpp" line="604"/>
+        <source>Format card as FAT32</source>
+        <translation type="unfinished">將記憶卡進行 FAT32 格式化</translation>
+    </message>
+    <message>
+        <location filename="../imagewriter.cpp" line="610"/>
+        <source>Use custom</source>
+        <translation type="unfinished">使用自訂映像檔</translation>
+    </message>
+    <message>
+        <location filename="../imagewriter.cpp" line="611"/>
+        <source>Select a custom .img from your computer</source>
+        <translation type="unfinished">從您的電腦選擇一個自訂的 .img 檔案</translation>
+    </message>
+    <message>
+        <location filename="../imagewriter.cpp" line="661"/>
         <source>Downloading and writing image</source>
         <translation>正在下載並寫入映像檔</translation>
     </message>
     <message>
-        <location filename="../imagewriter.cpp" line="575"/>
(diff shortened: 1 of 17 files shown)


src/i18n/rpi-imager_zh-TW.ts
```diff
@@ -467,277 +484,275 @@
...
         <source>Online - %1 GB download</source>
         <translation>線上 - %1 GB 下載</translation>
     </message>
     <message>
-        <location filename="../main.qml" line="833"/>
-        <location filename="../main.qml" line="885"/>
-        <location filename="../main.qml" line="891"/>
+        <location filename="../main.qml" line="899"/>
+        <source>Online - %1 MB download</source>
+        <translation type="unfinished"></translation>
```
tdewey-rpi, line 639: Perhaps this could copy-paste from the message above, substituting GB for MB?

src/i18n/rpi-imager_zh.ts
```diff
@@ -241,27 +241,47 @@
...
     <message>
-        <location filename="../imagewriter.cpp" line="442"/>
+        <location filename="../imagewriter.cpp" line="603"/>
+        <source>Erase</source>
+        <translation type="unfinished">擦除</translation>
+    </message>
+    <message>
+        <location filename="../imagewriter.cpp" line="604"/>
+        <source>Format card as FAT32</source>
+        <translation type="unfinished">将SD卡格式化为FAT32格式</translation>
```
tdewey-rpi, line 261: This doesn't look right - it's marked as "Unfinished", but definitely has a translation

src/i18n/rpi-imager_zh.ts
```diff
@@ -511,277 +524,267 @@
...
         <source>Online - %1 GB download</source>
         <translation>需要下载：%1 GB</translation>
     </message>
     <message>
-        <location filename="../main.qml" line="833"/>
-        <location filename="../main.qml" line="885"/>
-        <location filename="../main.qml" line="891"/>
+        <location filename="../main.qml" line="899"/>
+        <source>Online - %1 MB download</source>
+        <translation type="unfinished"></translation>
```
tdewey-rpi, line 671: As in zh-TW, maybe adapt the message above?

src/main.qml
```diff
@@ -894,7 +894,9 @@ ApplicationWindow {
                                 ? qsTr("Cached on your computer")
                                 : url.startsWith("file://")
                                   ? qsTr("Local file")
-                                  : qsTr("Online - %1 GB download").arg((image_download_size/1073741824).toFixed(1))
+                                  : image_download_size >= 1073741824
+                                  ? qsTr("Online - %1 GB download").arg((image_download_size/1073741824).toFixed(1))
```
tdewey-rpi, line 898: Thinking on this further - given we're reusing GB/MB across languages, perhaps we could make GB/MB and untranslated parameter to this string, and just prune it from the translations?

tdewey-rpi:
Conceptually, approved.

Implementation-wise, however, I think it'd be simpler to just extract the GB/MB specifier and make that an untranslated argument

src/i18n/rpi-imager_zh.ts
AaronDewes: It was like that previously. Probably the translator forgot to check the "finished" checkbox in Qt Linguist.

AaronDewes:
> Conceptually, approved.
> 
> Implementation-wise, however, I think it'd be simpler to just extract the GB/MB specifier and make that an untranslated argument

Yeah, that sounds better 👍

lurch:
Is it safe to assume that "GB" and "MB" are valid abbreviations for "GigaByte" and "MegaByte" in all foreign languages? E.g. IIRC French uses "MO" instead of "MB"? https://forums.commentcamarche.net/forum/affich-1921093-mb-en-mo-conversion-et-difference

tdewey-rpi:
> Is it safe to assume that "GB" and "MB" are valid abbreviations for "GigaByte" and "MegaByte" in all foreign languages? E.g. IIRC French uses "MO" instead of "MB"? https://forums.commentcamarche.net/forum/affich-1921093-mb-en-mo-conversion-et-difference
> 
> EDIT: See e.g. https://github.com/raspberrypi/rpi-imager/blob/qml/src/i18n/rpi-imager_fr.ts#L81-L82

Every day is a learning day!

I had assumed that since 'byte' is a technical term, it would be conserved across languages. Clearly, that was an incorrect assumption, and I would hate to impose an Anglosphere choice on the French-speaking world.

You've freedom of implementation, @AaronDewes, but I would _strongly_ prefer that we copied over the skeleton of the GB text translations where possible - translations are offered on a best-effort basis, and I'm confident the translation suppliers would welcome you adapting their translation to cover this subtly different presentation case.

tdewey-rpi:
I defer to @AaronDewes, but also suggest that could be a follow-up PR at a later date.

tdewey-rpi:
Implemented on 2.0 branch.

PR #753, opened 2023-12-15
qml: Clear saved password using onPressed handler

### LINKED ISSUE #531: QML Text Inputs drop first character of password
I have a Pi 4B and I have reinstalled the latest OS 64 Full onto my SD card using the Imager 1.7.3.
I have used the advanced settings to set up wireless, SSH login information. I set a custom name and password.
I have verified that the Pi is connecting to my wireless network.
I use Putty to try and log in. After inputting the username, I enter the password and all I get back is access denied.
I was looking at one of your website pages "https://www.raspberrypi.com/news/raspberry-pi-bullseye-update-april-2022/". And on that page, it says to add a text file. But as you keep reading it says you need to encrypt your password by running a certain program on your Pi. Well, if I cannot log onto my Pi how am I supposed to run a program on it to encrypt my password to allow me to log on.
Please see attached document to see all of the advanced settings, see my SD card image, error message logging in, and the section I mentioned regarding setting up a text file for your log in info.

Thank you for all of your help.

[Can't Log into my Pi 01-03-2023.pdf](https://github.com/raspberrypi/rpi-imager/files/10335582/Can.t.Log.into.my.Pi.01-03-2023.pdf)

tdewey-rpi:
Using onTextEdited meant that the clear() call would happen after the edit had actually happened.

Instead, use the Keys.onPressed handler, which happens earlier in the input chain.

Resolves #531

### DIFF
--- a/src/OptionsPopup.qml
+++ b/src/OptionsPopup.qml
@@ -153,7 +153,8 @@ Window {
                                     }
                                     TextField {
                                         id: fieldUserPassword
-                                        echoMode: TextInput.PasswordEchoOnEdit
+                                        echoMode: TextInput.Password
+                                        passwordMaskDelay: 2000 //ms
                                         Layout.minimumWidth: 200
                                         selectByMouse: true
                                         property bool alreadyCrypted: false
--- a/src/OptionsPopup.qml
+++ b/src/OptionsPopup.qml
@@ -153,19 +153,26 @@ Window {
                                     }
                                     TextField {
                                         id: fieldUserPassword
-                                        echoMode: TextInput.Password
+                                        echoMode: TextInput.PasswordEchoOnEdit
                                         Layout.minimumWidth: 200
                                         selectByMouse: true
                                         property bool alreadyCrypted: false
                                         property bool indicateError: false
 
-                                        onTextEdited: {
+                                        Keys.onPressed: (event)=> {
                                             if (alreadyCrypted) {
                                                 /* User is trying to edit saved
                                                    (crypted) password, clear field */
                                                 alreadyCrypted = false
                                                 clear()
                                             }
+
+                                            // Do not mark the event as accepted, so that it may
+                                            // propagate down to the underlying TextField.
+                                            event.accepted = false
+                                        }
+
+                                        onTextEdited: {
                                             if (indicateError) {
                                                 indicateError = false
                                             }


tdewey-rpi:
I've had out-of-band feedback that this patch drops security to an unacceptable level, and want to address that in the open.

In short, I don't believe there's a 'perfect' solution to entering passwords - every mechanism includes some degree of risk, either in terms of user convenience or information leakage. I'd like to address two specific points that were raised:

### What if someone looks over the user's shoulder at the screen?

The particular case raised here was one of a busy classroom, and I absolutely consider it a fair comment.

That said, I don't believe that showing the password during entry is a materially worse state than already exists - where the malicious user could simply look at the keyboard being used instead. I consider it extremely unlikely that a user will enter a password in a situation where an attacker isn't in the position to also view their keypresses - the obvious counterexample being demonstration videos - where I wouldn't expect the demonstrator to be using an actual hardened credential anyway!

### Showing the password permanently on screen is wrong

I agree - which is why this mechanism only shows it during entry, where the user can review and confirm their input. Unlike other solutions, this mechanism does provide a mechanism for leaking the password post-entry, which is why I'm considering also migrating the WiFi password field to use this mechanism too. Today, using the 'show password' semantic, a malicious user with post-input control of the device can open Raspberry Pi Imager and potentially obtain someone else's WiFi credentials.

To further illustrate how this change behaves, I've recorded a video.

https://github.com/raspberrypi/rpi-imager/assets/146750643/9175a0b8-f7b4-43cc-ac10-1eab8d7db7ae

tdewey-rpi:
Additionally, I'd like to leave a note that I considered the alternative of leaving the password field to mask as it does today, but extending the mask time to allow you to at least see the last character.

While the more conventional solution, I don't believe it would help people providing a password from a password manager - as even when pasting contents it would still only leave the last character unmasked for the period selected - relying on a user to count the asterisks to determine if the length was correct.

Such a system would also have the same problem of over-the-shoulder malicious observers - this time they'd just observe the characters as they were typed, or copied from a password generator. The only flow where I'd see this having an advantage was one where you generated the password in a secure place, stored it in a password manager, copied the password unseen at a 'busy' location, and pasted it into the password field - revealing only the last character.

I consider that scenario to be somewhat convoluted - not least because it would require the user to have not saved their password as part of a customisation flow a priori, but instead to customise on the spot in an untrusted location.

lurch:
> stored it in a password manager...

Also, presumably people will be typing passwords into RPi Imager _much_ less often than they log into Facebook or their online banking :grinning:

tdewey-rpi:
Changed to use character-dwelling mechanism, following a potential information leak when you used 'undo' on the TextField - which in testing, produced partially correct data.

cillian64 (approved):
lgtm

tdewey-rpi:
Merging on that basis - this impacts a surprisingly large number of people.

cillian64:
(this has the slightly odd effect that highlighting the password field then pressing "ctrl" clears the password, but pressing Ctrl+Z restores the crypted password so that seems ok.  And because it's the proper password box, restoring the password doesn't reveal anything.

Also, I kinda expected to be able to reveal an obscured password by backspacing and then undo-ing to reveal each character using the dwell time.  But it seems like Qt thought of that and characters are not revealed in this case).

PR #771, opened 2024-01-05
Updated Branding, URLs

tdewey-rpi:
raspberrypi.org is owned by the Raspberry Pi Foundation, and for cleanliness, we should point instead to raspberrypi.com (which is run by Raspberry Pi Ltd) instead.

Additionally, let's make sure the application name reflects the...application name! "Imager" was unfortunately a little too short to distinguish from tools included in various Linux distros, and avoiding confusion is the aim.

### DIFF
--- a/debian/changelog
+++ b/debian/changelog
@@ -1,3 +1,12 @@
+rpi-imager (1.8.5) unstable; urgency=medium
+
+  * Downloader: Fixed #767, which caused terminated downloads
+                            when connections stalled for >60s
+  * Branding: Update the URL, application name to match branding
+              guidelines.
+
+ -- Floris Bos <bos@je-eigen-domein.nl>  Wed, 10 Jan 2024 16:00:00 +0200
+
 rpi-imager (1.8.4) unstable; urgency=medium
 
   * OS customisation: Fixed #531, which caused password entry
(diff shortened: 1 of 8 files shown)


tdewey-rpi:
Not safe to merge yet - turns out QSettings uses the name as the config file name, so we'll need a way to either compel QSettings back to using the previous name or a migration mechanism that is reliable.

maxnet:
There may be other places where .org and the original application name (imagingutility) is used, like "mac/Info.plist.in"

I am always a bit hesitant with changes such things though...
In addition to saved configuration settings no longer working, changing application identifiers is likely to void system settings like file associations, granted permissions (whether or not the program is allowed to access removable media on Mac) and may affect any white-listing we may have in anti-virus/corporate security programs.

cillian64 (approved):
Agreed that some care is needed, but the name change in the linux .desktop file shouldn't affect file associations, and we're not changing the executable name or config file name/location.

maxnet:
>Agreed that some care is needed, but the name change in the linux .desktop file shouldn't affect file associations

Wasn't too concerned about the .desktop file.

```
    _app->setOrganizationDomain("raspberrypi.com");
    _app->setApplicationName("Raspberry Pi Imager");
```

But for example the application name set here is used at several less obvious places like in DBus communication.

If you also would change the Mac plist it do could affect things.

tdewey-rpi:
Dropping, this will come with 2.0

PR #772, opened 2024-01-05
imagewriter: Force permissions on settings to 600

### LINKED ISSUE #770: [BUG] Insecure permissions for configuration file holding hashed password & plaintext WiFi password
**Describe the bug**

The configuration file which holds the hashed & plaintext WiFi has insecure permissions by default.

**To Reproduce**

on Linux:

```
$ ls -l "$HOME/.config/Raspberry Pi/Imager.conf"
-rw-rw-r--. 1 ricardo ricardo 775 Nov 19 12:13 Imager.conf
$ grep Password "$HOME/.config/Raspberry Pi/Imager.conf"
sshUserPassword=$5$xxx
wifiPassword=xxx
```

**Expected behaviour**

0600 permissions.

tdewey-rpi:
Use this mechanism to replace an older scheme that fully read and then fully wrote the file.

Resolves #770

### DIFF
--- a/src/imagewriter.cpp
+++ b/src/imagewriter.cpp
@@ -103,30 +103,15 @@ ImageWriter::ImageWriter(QObject *parent)
 #ifdef Q_OS_WIN
     _taskbarButton = nullptr;
 #endif
-
-    if (!_settings.isWritable() && !_settings.fileName().isEmpty())
     {
-        /* Settings file is not writable, probably run by root previously */
-        QString settingsFile = _settings.fileName();
-        qDebug() << "Settings file" << settingsFile << "not writable. Recreating it";
-        QFile f(_settings.fileName());
-        QByteArray oldsettings;
+        const auto expectedPermissions = (QFileDevice::WriteOwner | QFileDevice::ReadOwner);
+        auto settingsFile = QFile(_settings.fileName());
 
-        if (f.open(f.ReadOnly))
-        {
-            oldsettings = f.readAll();
-            f.close();
-        }
-        f.remove();
-        if (f.open(f.WriteOnly))
-        {
-            f.write(oldsettings);
-            f.close();
-            _settings.sync();
-        }
-        else
-        {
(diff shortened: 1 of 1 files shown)


maxnet:
Upon information and belief QSettings does not create any file on disk until you either call sync() or the QSettings object is destructed.


You are now correcting permission on Imager's start (ImageWriter's constructor), if the .conf file exists at that time.

So let's imagine that I have never used Imager before (so .conf does not exist yet), use it to write the single SD card I have, and then won't be using Imager for a while.
What permissions will the .conf in my home directory have?


Also be aware that on Windows the fileName() returned will be a registry key.
While you are probably saved by your file existence check, you may wish to #ifndef Q_OS_WIN the code just in case.

tdewey-rpi:
A fair comment - and I'll see if we can compel QSettings into creating a file with the appropriate permissions.

maxnet:
>A fair comment - and I'll see if we can compel QSettings into creating a file with the appropriate permissions.

Think that it is not doing that, may actually be a long term Qt bug.

https://github.com/qt/qtbase/blob/dev/src/corelib/io/qsettings.cpp#L1509-L1513

Don't think the "fileInfo.permissions() |" part is supposed to be there.
But getting such things fixed upstream may be problematic. Change in behavior, etc.

tdewey-rpi:
Agreed - this looks like an upstream issue. I've just had a test of the scenario you describe - and sure enough after a `sync()` we get a 644 file.

The only comprehensive fix I can see with the QSettings API is to register our own read & write functions, and then re-implement the body of the QSettings file reader and writer. Deeply undesirable, but might be the only way to coerce it into doing the right thing.

https://doc.qt.io/qt-6/qsettings.html#registerFormat

tdewey-rpi:
@cillian64 If you have bandwidth, might be worth looking at this horror as mentioned earlier. Ignoring that it's a TOC/TOU problem, in the near-term we can probably address it by bunding a setPermissions(...) call on the back of every flush().

PR #790, opened 2024-01-23
README.md: Add rebuild instructions for Arch based distros

inferno0230:
Tests:
Rebuild rpi-imager and flashed ubuntu-server onto a sdcard for rpi 5

### DIFF
--- a/README.md
+++ b/README.md
@@ -89,6 +89,33 @@ make
 sudo make install
 ```
 
+### Arch/Manjaro Linux
+
+#### Get dependencies
+
+Install the build dependencies:
+
+```
+sudo pacman -S git gcc make cmake libarchive curl openssl qt5-base qt5-quickcontrols2 qt5-svg qt5-tools qt5-x11extras xz
+```
+
+#### Get the source
+
+```
+git clone --depth 1 https://github.com/raspberrypi/rpi-imager
+```
+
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -89,6 +89,33 @@ make
...
+
+```
+cd rpi-imager
+mkdir -p build
+cd build
+cmake ../src
+make
+sudo make install
+```
+
```
tdewey-rpi, line 118: This section is superfluous and repetitive - it's a direct clone of wording ~10 lines above. In your future changes, please consider conciseness as an imperative in documentation, as a favour to both the end user and the reviewer.

tdewey-rpi:
Thanks for the submission, @inferno0230.

Unfortunately I'm not minded to accept this change:

1. We do not offer explicit development support for distros that regular maintainers do not use. As an example, I develop on Raspberry Pi OS, macOS and Windows.
2. The change is unnecessarily large - even if we were to merge Arch build instructions, they would need to be constrained to the bare minimum of instruction change to get Arch users on board - which as far as I can tell boils down to a call to `pacman` with the different package names of the Arch ecumene.

Please don't let this put you off suggesting other changes - but please be aware that documentation changes in particular are strongly considered along the axes of clarity, conciseness and correctness.

PR #792, opened 2024-01-26
appstream metainfo: Fix appstream-builder generation

joebonrichie:
launchable desktop-id type was no longer matching after 06abbadbb501869c29076df0f0409b977869e992 causing appstream-builder to fail to generate appdata for this package.


tdewey-rpi:
Thanks for the submission, @joebonrichie 

And great catch!

PR #805, opened 2024-02-06
imagewriter: Use QCoreApplication::translate

### LINKED ISSUE #799: [BUG] C++ tr() strings are not translated
**Describe the bug**
Translated strings in C++ are not translated

**To Reproduce**
1. Install Raspberry Pi Imager v1.8.5
2. Press (CMD + SHIFT + L)
3. Select any non-English language
4. Open 'Choose OS'
5. Scroll to the bottom of the list

**Expected behaviour**
'Erase' and 'Use Custom' to have translated labels

**Screenshots or video**
Screenshot 2024-01-19 092636

**Desktop (please complete the following information):**
 - OS: macOS Sonoma, Windows, Raspberry Pi OS

**Additional context**
This only appears to affect strings using the tr() macro in the C++ side of the codebase. The QML side translates dynamically as expected.

tdewey-rpi:
tr() is not safe for strings that you move between contexts, as it fails to provide a mechanism for specifying the context of the string.

Regenerate the translation files, and use QCoreApplication to locate the real translations.

Additionally, bodge a race condition between translator replacement and OS list fetching, by marking the OS list as something to be re-calculated on a translator change.

Fixes #799

### DIFF
--- a/src/i18n/rpi-imager_ca.ts
+++ b/src/i18n/rpi-imager_ca.ts
@@ -97,59 +97,59 @@
         <translation>S&apos;està iniciant la baixada</translation>
     </message>
     <message>
-        <location filename="../downloadthread.cpp" line="466"/>
+        <location filename="../downloadthread.cpp" line="468"/>
         <source>Error downloading: %1</source>
         <translation>S&apos;ha produït un error en la baixada: %1</translation>
     </message>
     <message>
-        <location filename="../downloadthread.cpp" line="663"/>
+        <location filename="../downloadthread.cpp" line="665"/>
         <source>Access denied error while writing file to disk.</source>
         <translation>S&apos;ha produït un error d&apos;accés denegat en escriure el fitxer al disc.</translation>
     </message>
     <message>
-        <location filename="../downloadthread.cpp" line="668"/>
+        <location filename="../downloadthread.cpp" line="670"/>
         <source>Controlled Folder Access seems to be enabled. Please add both rpi-imager.exe and fat32format.exe to the list of allowed apps and try again.</source>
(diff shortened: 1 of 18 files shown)


lurch:
Apologies if this is a stupid question, but how do you identify which strings "you move between contexts" ? (and therefore where you need to use `QApplication::translate` rather than just `tr` ?)

tdewey-rpi:
> Apologies if this is a stupid question, but how do you identify which strings "you move between contexts" ? (and therefore where you need to use `QApplication::translate` rather than just `tr` ?)

Unfortunately as I've mentioned the tooling doesn't help here. It's down to the individual developer to notice they are moving a translatable string, and annotate the new location with the previous context.

lurch:
Ahhh, I misunderstood - I thought you were talking about the strings moving between different contexts at runtime, rather than simply the translatable strings being moved to a different "context" (i.e. file) in the source-code :slightly_smiling_face:

tdewey-rpi:
> Ahhh, I misunderstood - I thought you were talking about the strings moving between different contexts at runtime, rather than simply the translatable strings being moved to a different "context" (i.e. file) in the source-code 🙂

Were it so simple! Contexts are logical units inside the source tree - typically tied to a QObject name, but I can't see any reason you couldn't have arbitrary ones. Moving things to arbitrary named contexts might prove more durable longer-term.

lurch:
> Moving things to arbitrary named contexts might prove more durable longer-term.

I guess that would also allow you to re-organise the translatable strings into groupings that make more (logical) sense for the people doing the translations, rather than simply being grouped depending on where those strings appear in the source code? Does the qml-side of this whole translation-shenanigans also allow the use of arbitrary named contexts?
(although for a relatively simple application like RPi Imager, there might not be enough translatable text to warrant re-organising the translations. Sorry, went off on a bit of a side-tangent...)

tdewey-rpi:
Yes, that's one of the advantages I can see.

qml does offer a similar context-selecting translation function: https://doc.qt.io/qt-6/qml-qtqml-qt.html#qsTranslate-method

cillian64 (approved):

Issue #757, opened 2023-12-20
[BUG] Order of YES and NO buttons differs between popups, causing confusion.


SkyLined:
**Describe the bug**
* The [MsgPopup.qml](https://github.com/raspberrypi/rpi-imager/blob/qml/src/MsgPopup.qml) file has a [NO](https://github.com/raspberrypi/rpi-imager/blob/qml/src/MsgPopup.qml#L98) and a [YES](https://github.com/raspberrypi/rpi-imager/blob/qml/src/MsgPopup.qml#L107) button, in that order.
* The [UseSavedSettingsPopup.qml](https://github.com/raspberrypi/rpi-imager/blob/qml/src/UseSavedSettingsPopup.qml) file has a [YES](https://github.com/raspberrypi/rpi-imager/blob/qml/src/UseSavedSettingsPopup.qml#L125) and a [NO](https://github.com/raspberrypi/rpi-imager/blob/qml/src/UseSavedSettingsPopup.qml#L134) button, in that order.

The different order makes it hard to step through these popups if they are shown in succession; if you see YES and NO in one popup, you expect the same order in the next. I've repeatedly clicked NO in the second popup even though I wanted to click YES because I expected the buttons to be in the same order.

**To Reproduce**
Write Raspberry Pi OS to a device; the imager will ask "Would you like to apply OS customization settings?" using the `UseSavedSettingsPopup.qml` template (order = YES, NO). If you choose YES or NO, it will ask you "Are you sure you want to continue?" using the `MsgPopup.qml` template (order = NO, YES).

**Expected behaviour**
I expect consistency in the order of GUI components across the application.

**Screenshots or video**
![imager1](https://github.com/raspberrypi/rpi-imager/assets/1959110/1f391aa2-0b93-4960-bbab-36728ac7ee1b)
![imager2](https://github.com/raspberrypi/rpi-imager/assets/1959110/aff5c22f-6d92-43e6-80c7-82729e945273)

**Desktop:**
 - OS: Windows 11 (but should be irrelevant)

**The name of the OS you are trying to write**
- Any Raspberry PI OS (but should be irrelevant)

**Are you using OS Customisation?**
No ( (but should be irrelevant)

**Additional context**
N/A

lurch:
See also #624

tdewey-rpi:
Thanks for the report, @SkyLined.

While reordering the buttons would be the most straightforward route, I suspect the better option is to actually redesign both dialogs.

[...]

PR #807, opened 2024-02-06
Swap order of "yes" and "no" buttons in OS customisation popup

sea-witch:
Swaps the order of the "yes" and "no" button in the OS customisation popup, making this popup conform to the convention of keeping the _confirmation_ option to the right.

Resolves https://github.com/raspberrypi/rpi-imager/issues/757

Screenshot 2024-02-06 at 3 28 00 PM

### DIFF
--- a/src/UseSavedSettingsPopup.qml
+++ b/src/UseSavedSettingsPopup.qml
@@ -121,6 +121,14 @@ Popup {
             }
 
             ImButtonRed {
+                text: qsTr("NO")
+                onClicked: {
+                    msgpopup.close()
+                    msgpopup.no()
+                }
+            }
+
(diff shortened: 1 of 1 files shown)


tdewey-rpi:
The current ordering is designed to present the path of least surprise if you just keep clicking the rightmost-option.

I very deliberately did _not_ order the buttons this way, as under this flow you'll get whatever was stored in your settings. Furthermore, if you _don't_ have anything stored in settings, the rightmost button is now unavailable.

Rejecting this PR, as I don't believe this solution will present a safer default action. I believe the only concrete way to address this - as I've stated elsewhere - is to introduce a different UI model for this action.

PR #816, opened 2024-02-09
Exposing 'DRIVELIST_FILTER_SYSTEM_DRIVES' as a cmake variable

### LINKED ISSUE #534: Feature Request: expose 'DRIVELIST_FILTER_SYSTEM_DRIVES' as a cmake var
I personally use rpi-imager with `DRIVELIST_FILTER_SYSTEM_DRIVES` in config.h set to false, because I need to write to SATA drives, so I was wondering whether it would be beneficial to expose it as a cmake arg.

I already created a [commit in my fork](https://github.com/MrSmoer/rpi-imager/commit/4074bcab2fd007b1845ae0d0b97578d08c22410c), should I open a PR? 

What do you think?
Thank you, have a nice day!

MrSmoer:
Makes the filtering-behaviour userconfigureable during compile-time. This way its easier to configure the imager for flashing SATA or NVME Drives installed in your system (because of lack of adapters), or if the drives are (accidentally) labled as systemdrives by drivelist.

By default system drives are still filtered.

Fixes #534

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -4,6 +4,7 @@
 cmake_minimum_required(VERSION 3.9.4)
 OPTION (ENABLE_CHECK_VERSION "Check for version updates" ON)
 OPTION (ENABLE_TELEMETRY "Enable sending telemetry" ON)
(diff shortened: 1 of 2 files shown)


tdewey-rpi:
Awesome, thanks @MrSmoer!

PR #819, opened 2024-02-14
OS Customisation: Validate usernames

tdewey-rpi:
Import the systemd recommended username validation regex.

Resolves a ticket reported to the Raspberry Pi information inbox.


tdewey-rpi:
Tested on Raspberry Pi OS - this will now refuse any uppercase characters in the username, and constrain the length to a value that should work on any distro that runs systemd.

lurch:
See also #741

PR #832, opened 2024-03-05
imagewriter: Disable customisation of custom IMGs

### LINKED ISSUE #831: [BUG]: Existing users are overwritten when applying custom settings
### What happened?

1. I have an image with custom users pre-defined.
2. I flash the image to the an SD card using custom settings (wifi / hostname and base user)
3. The existing user was overwritten by those settings
4. The image, which relies on that deleted user is now not working as intended

### Version

1.8.5 (Default)

macOS

### Host OS Version

12.6.2 (21G320)

### Selected OS

Custom Talkkonnect image

Raspberry Pi Zero, Zero W, Zero WH

microSD Card in a USB reader

### OS Customisation

### Relevant log output

_No response_

tdewey-rpi:
Resolves #831 by not offering customisation for custom image sources.

We have no associated metadata for these images, so customisation isn't something we can offer in a clear and easy to understand way.

### DIFF
--- a/src/imagewriter.cpp
+++ b/src/imagewriter.cpp
@@ -1293,6 +1293,10 @@ void ImageWriter::clearSavedCustomizationSettings()
     _settings.sync();
 }
 
+void ImageWriter::disableImageCustomization() {
+    _initFormat.clear();
+}
+
 bool ImageWriter::hasSavedCustomizationSettings()
 {
     _settings.sync();
--- a/src/imagewriter.h
+++ b/src/imagewriter.h
@@ -124,6 +124,7 @@ public:
 
     Q_INVOKABLE bool getBoolSetting(const QString &key);
     Q_INVOKABLE void setSetting(const QString &key, const QVariant &value);
+    Q_INVOKABLE void disableImageCustomization();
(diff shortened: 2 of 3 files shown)


XECDesign:
I can't be the only one who frequently flashes Raspberry Pi OS from previously downloaded images. If somebody flashes a random image and find that the customisations don't apply, I think the problem is with that image rather than the imager.

tdewey-rpi:
While I would generally agree, @XECDesign, the problem here is that we can't communicate that failure to the user in a clear and concise manner.

As #831 demonstrates, users will blame `rpi-imager` even when the image creator has admitted a fault in their guide or their set up.

Let me talk to you offline about a way around this particular issue that I think you might appreciate.

cillian64:
I'm with Serge on this one, especially because image files are the only way to keep a consistent version and are the only option when working offline.

I do think it would be useful to have a warning message when applying customisation to image files that results may be unexpected if it's not an rpi-os image.  There will probably be occasional bug reports nonetheless but I reckon the complaints are outweighed by the people relying on this.

If this does really need solving, could we come up with a simple way to identify whether image files are rpi-os, ubuntu server, or something else? (a way which doesn't involve adding a new identifier and so works on old image files).

tdewey-rpi:
Discussed offline, I'm going to withdraw this and submit an alternative.

The alternative will, rather than completely removing customisation, provide an alternative additional dialog that asks you to confirm the OS is either Raspberry Pi OS or Ubuntu Server. If you select neither, customisation will _not_ be offered.

PR #860, opened 2024-04-30
cli: Add disable-eject

### LINKED ISSUE #851: [FEATURE]: Add disable-eject to CLI
### Is your feature request related to a problem? Please describe.

I'm using the CLI to write a .img to a SSD but I need to make further modifications after writing the image, namely create new partitions and increase partitions size. If I use the GUI version this option is available but I cannot make it work in the CLI version.

### Describe the solution you would like to see implemented

Add --disable-eject (or similar) to the CLI options.

### Describe alternatives you've considered

I've tried to use the GUI versions and it all works fine, but when I try to play with partitions after using the CLI version it does not work, I need to unplug and plug back in the disk.

### Additional context

.

### Version

1.8.5 (Default)

tdewey-rpi:
Add a new CLI flag, `--disable-eject`, which maps to the option in the UI to not automatically eject the storage media after writing.

While we're here, use a more concise declaration form for the cli options.

Resolves #851

### DIFF
--- a/src/cli.cpp
+++ b/src/cli.cpp
@@ -49,27 +49,19 @@ Cli::~Cli()
 int Cli::main()
 {
     QCommandLineParser parser;
-    QCommandLineOption cli("cli");
-    parser.addOption(cli);
-    QCommandLineOption disableVerify("disable-verify", "Disable verification");
-    parser.addOption(disableVerify);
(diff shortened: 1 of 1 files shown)


tdewey-rpi:
Requires testing on Linux before merge.

cillian64 (approved):
Haven't tested (sounds like you have) but the code change looks good to me.

PR #867, opened 2024-05-09
localectl doesn't work on modern Debian/Ubuntu

waveform80:
On modern Debian and Ubuntu releases (23.10 onwards for Ubuntu), `localectl` does not (currently) operate [1] [2]. The `keyboard:` section should be used instead, which should do the right thing on whatever platform cloud-init is being used on (writes `/etc/default/keyboard` on Debian/Ubuntu, uses `localectl` on RHEL/Fedora, runs `setup-keymap` on Alpine).

[1]: https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=1038762

[2]: https://bugs.launchpad.net/ubuntu/+source/cloud-init/+bug/2030788

### DIFF
--- a/src/OptionsPopup.qml
+++ b/src/OptionsPopup.qml
@@ -787,8 +787,9 @@ Window {
             addFirstRun("fi")
 
             addCloudInit("timezone: "+fieldTimezone.editText)
-            addCloudInitRun("localectl set-x11-keymap \""+fieldKeyboardLayout.editText+"\" pc105")
-            addCloudInitRun("setupcon -k --force || true")
+            addCloudInit("keyboard:")
+            addCloudInit("  model: pc105")
+            addCloudInit("  layout: \"" + fieldKeyboardLayout.editText + "\"")
         }
 
         if (firstrun.length) {


maxnet:
>The keyboard: section should be used instead

Do all Ubuntu editions you offer through Imager support the keyboard module?

(I think I contributed the keyboard module to cloud-init in 2022.
So I can imagine Ubuntu 20 lacking it, unless it has been back ported).

waveform80:
Good question -- I know jammy and noble both do, but you're correct that focal's still in the list (though if we're going to break keyboard on a release, I'd rather break focal than noble). Still, cloud-init is one of those things that is constantly back-ported, but it'll depend if the current point-release of focal was spun before / after the release of cloud-init that added it.

Looking at the manifest of the current point release of focal (.5), it's got cloud-init 22.2. And ... *(goes digging in the cloud-init repo)*, yes! It's got the keyboard module, so we should be good.

> (I think I contributed the keyboard module to cloud-init in 2022. So I can imagine Ubuntu 20 lacking it, unless it has been back ported).

So [you did](https://bugs.launchpad.net/cloud-init/+bug/1951593), thank you (being an aficionado of `ctrl:nocaps`, it's something I constantly use, so I very much appreciate it)! Ooh, 22.1 -- so that *just* sneaked into SRU in time for focal's .5 release, nice.

tdewey-rpi:
> It's got the keyboard module, so we should be good.

Merging on the basis of contributor testing and verification.

Thanks for the submission, @waveform80

PR #876, opened 2024-06-01
Switch YES and NO button to make order consistent across the application

SkyLined:
The order of YES and NO buttons differs between windows, causing confusion, as reported in https://github.com/raspberrypi/rpi-imager/issues/757.

There are plans for a more perfect fix but those plans has not been acted on in months. I'd rather have a good fix now than continue to press the wrong button until somebody has time to implement that.

### DIFF
--- a/src/MsgPopup.qml
+++ b/src/MsgPopup.qml
@@ -95,21 +95,21 @@ Popup {
             spacing: 20
 
             ImButtonRed {
-                text: qsTr("NO")
-                onClicked: {
-                    msgpopup.close()
-                    msgpopup.no()
-                }
-                visible: msgpopup.noButton
-            }
-
-            ImButtonRed {
                 text: qsTr("YES")
                 onClicked: {
                     msgpopup.close()
                     msgpopup.yes()
                 }
                 visible: msgpopup.yesButton
+            }
+
+            ImButtonRed {
+                text: qsTr("NO")
+                onClicked: {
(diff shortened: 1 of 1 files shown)


tdewey-rpi:
In #757, I explained why I considered this current order to be the safest inside of this UI model:

> If the argument was not one that would potentially present a user with a surprising OS configuration, one that deviates from the documentation of the OS vendor, then I may take a more charitable view of the proposal.

> However, that is not what is being proposed here. If you press 'no', you're getting the OS as the OS provider intended. If you press 'yes' and are not actively aware of the customisations you have configured you will get an OS that is configured in a manner you do not expect - which is potentially much more harmful than going through the OOBE of an OS.

I am, of course, prepared to be wrong. So I instead looked at the number of issues raised against this, and I found....3, with no additional upvotes.

Further, at the end of my comment on #757, I invited interested parties to propose a design for how they'd expect this box to work - one that fundamentally altered it's model to put greater emphasis on having the user make a _deliberate_ choice:

> If you wish to make a proposal for that UI redesign, please raise a Feature Request.

Unfortunately, this PR therefore:

1. Appears to affect, at most, 3 individuals.
2. Would change the 'default' path to one with potentially harmful behaviour
3. Did not take up my invitation to consider this function in the round

As a result, I'm rejecting this PR, and again inviting interested parties to consider this function in the round.

PR #895, opened 2024-07-16
Fix GPT signature test

jadonk:
gpt.Signature is not a null-terminated string, but an 8-byte character array. Using `strcmp` will not provide a valid result. Use `strncmp` over the first 8 bytes instead.


tdewey-rpi (approved):

tdewey-rpi:
Thanks for the submission, @jadonk!

PR #903, opened 2024-08-07
Linux/OSX: Fix reversed fsync check

ali1234:
This if() condition checks the result of seek, write, flush and fsync. The first three are methods on QFile, so return a truthy value on success. write() returns number of bytes written while seek() and flush() return bool true on success. Therefore they have to be inverted to detect errors. However, fsync() is a posix system call, so it returns zero (false) on success. Therefore it should not be inverted.

Note that currently this entire code block can never run. knownsize is always zero because QFile.size() currently cannot determine the size of a block device. However, when I asked to confirm this behavior, a Qt maintainer decided to implement it, so there is now a PR. If that is released then this code block will become live and the reversed fsync check will cause a problem.

PR: https://codereview.qt-project.org/c/qt/qtbase/+/581443

### DIFF
--- a/src/downloadthread.cpp
+++ b/src/downloadthread.cpp
@@ -314,7 +314,7 @@ bool DownloadThread::_openAndPrepareDevice()
         if (!_file.seek(knownsize-emptyMB.size())
                 || !_file.write(emptyMB.data(), emptyMB.size())
                 || !_file.flush()
-                || !::fsync(_file.handle()))
+                || ::fsync(_file.handle()))
         {
             emit error(tr("Write error while trying to zero out last part of card.<br>"
                           "Card could be advertising wrong capacity (possible counterfeit)."));


tdewey-rpi:
Thanks for the submission, @ali1234

lurch:
> knownsize is always zero because QFile.size() currently cannot determine the size of a block device. However, when I asked to confirm this behavior, a Qt maintainer decided to implement it, so there is now a PR. If that is released then this code block will become live and the reversed fsync check will cause a problem.

It might be a while until a version of QT containing that fix filters down to RPi Imager? In the meantime, I think the `drivelist` code might report the exact size of the drive's block device? (or maybe it only reports it to the nearest MB / GB, I can't remember)

maxnet:
>In the meantime, I think the drivelist code might report the exact size of the drive's block device? 

Wouldn't bother with that.

Just get the Unix file descriptor of the QFile with handle() and perform the same ioctl() the Qt guys are going to implement....

PR #917, opened 2024-09-09
Fix duplicate string in OS list

Kurtsley:
The "(Recommended)" string can be appended twice when the description already contains (Recommended) as is the case with the Pi OS 64 bit image:
image

This adds a check to ensure this doesn't happen.

### DIFF
--- a/src/main.qml
+++ b/src/main.qml
@@ -1619,7 +1619,9 @@ ApplicationWindow {
             var candidate = oslist_parsed[0]
 
             if ("description" in candidate && !("subitems" in candidate)) {
-                candidate["description"] += " (Recommended)"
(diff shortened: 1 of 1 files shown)


lurch:
Related to #694 and #703 (which imply that this is a "bug" in the JSON, rather than in Imager?) ping @tdewey-rpi

tdewey-rpi:
Thanks for the submission, @Kurtsley.

Unfortunately, I already have an identical change in my PR queue for 1.9.0, which I'm expecting to send into the main branch today. I agree that this is probably the best solution - while the obvious path is to remove the string from the JSON, this mechanism preserves compatibility with 1.8.5 and earlier releases.

Kurtsley:
Sounds good, I look forward to 1.9.0!

Issue #852, opened 2024-04-18
[FEATURE]: Add locale to the imager


knutegit:
### Is your feature request related to a problem? Please describe.

Image has timezone and wifi country but defaults to en_gb.

### Describe the solution you would like to see implemented

Add a locale selection to imager

### Describe alternatives you've considered

If I remember I can use raspi-config to fix the locale.

### Additional context

No

### Version

1.8.5 (Default)

lurch:
Imager already has locale settings - see the screenshots at https://www.raspberrypi.com/documentation/computers/getting-started.html#advanced-options

Are you flashing Raspberry Pi OS to your SD card, or some other OS?

tdewey-rpi:
Feature request rejected.

This is already a feature of Raspberry Pi Imager.

scruss:
> Imager already has locale settings ...

No, it doesn't. You *can* set the time zone and the keyboard language, but `$LANG` is always set to en_GB.UTF-8 (and similarly, the output of `locale` is mostly en_GB). To have a proper locale setting, you need to have the equivalent of
```
dpkg-reconfigure locales
```

See forum thread: [Raspberry Pi Imager improvement request - Raspberry Pi Forums](https://forums.raspberrypi.com/viewtopic.php?p=2220151#p2220151)

lurch:
Oh! Sorry that I misunderstood the request.

ZuiMelanieForno:
I use this Workaround.

In the `firstrun.sh` I added the following line before the deletion-command of the firstrun.sh
`/usr/bin/raspi-config nonint do_change_locale de_DE.UTF-8`

```bash

KBEOF
    dpkg-reconfigure -f noninteractive keyboard-configuration
fi

/usr/bin/raspi-config nonint do_change_locale de_DE.UTF-8

rm -f /boot/firstrun.sh
sed -i 's| systemd.run.*||g' /boot/cmdline.txt
exit 0

```

```bash
<some prompt> locale
LANG=de_DE.UTF-8
LANGUAGE=
LC_CTYPE="de_DE.UTF-8"
LC_NUMERIC="de_DE.UTF-8"
LC_TIME="de_DE.UTF-8"
LC_COLLATE="de_DE.UTF-8"
LC_MONETARY="de_DE.UTF-8"
LC_MESSAGES="de_DE.UTF-8"
LC_PAPER="de_DE.UTF-8"
LC_NAME="de_DE.UTF-8"
[...]

PR #919, opened 2024-09-11
Feat/add locale support

digitalLumberjack:
Hi, see https://github.com/raspberrypi/rpi-imager/issues/852

PR opened for discussion about what we could add for locales.

I added the support for setting the locale in OS Customization (and automatic detection of it based on the timezone, for GB, USA, and France).

The locales list (locales.txt) comes from:
```
cat /usr/share/i18n/SUPPORTED
```
on Raspberry Pi OS.

The ui renders like this:
image

The `firstuse.sh` scripts is modified to use `localectl set-locale LANG=$LOCALE LC_ALL=$LOCALE LANGUAGE=$LOCALE` on first boot.
(`localectl` use `locale-gen` and update the `/etc/default/locale` file

```
#!/bin/bash

set +e

if [ -f /usr/lib/raspberrypi-sys-mods/imager_custom ]; then
   /usr/lib/raspberrypi-sys-mods/imager_custom set_keymap 'fr'
   /usr/lib/raspberrypi-sys-mods/imager_custom set_timezone 'Europe/Paris'
else
   rm -f /etc/localtime
   echo "Europe/Paris" >/etc/timezone
   dpkg-reconfigure -f noninteractive tzdata
cat >/etc/default/keyboard <<'KBEOF'
XKBMODEL="pc105"
XKBLAYOUT="fr"
XKBVARIANT=""
XKBOPTIONS=""

KBEOF
   dpkg-reconfigure -f noninteractive keyboard-configuration
fi
localectl set-locale LANG=fr_FR.UTF-8 LC_ALL=fr_FR.UTF-8 LANGUAGE=fr_FR.UTF-8
rm -f /boot/firstrun.sh
sed -i 's| systemd.run.*||g' /boot/cmdline.txt
exit 0
```

The user_data cloudinit yaml has the new entry locale (see https://cloudinit.readthedocs.io/en/latest/reference/modules.html#locale):
```
timezone: Europe/Paris
keyboard:
  model: pc105
  layout: "fr"
locale: fr_FR.UTF-8
```

Open questions:
- How many locales would you like to keep in this list?
- Is the local file with `LOCALECODE [(Human Readable)]` something you would like to keep ?
- We could detect the host locale, but I was not able to get the full locale code from QLocale class for now. It seems to have been heavily simplified on Qt.
- Should we use the host Locale (if we can detect it) to preset the keyboard ?

Thanks to every contributor of this great tool :)

### DIFF
--- a/src/imagewriter.cpp
+++ b/src/imagewriter.cpp
@@ -1250,7 +1250,7 @@ void ImageWriter::setImageCustomization(const QByteArray &config, const QByteArr
 
     qDebug() << "Custom config.txt entries:" << config;
     qDebug() << "Custom cmdline.txt entries:" << cmdline;
-    qDebug() << "Custom firstuse.sh:" << firstrun;
+    qDebug() << "Custom firstrun.sh:" << firstrun;
     qDebug() << "Cloudinit:" << cloudinit;
 }
 


tdewey-rpi:
Thanks for the PR, @digitalLumberjack!

This is indeed a long-standing pain point, and I'm particularly impressed to see a member of the community stepping up to try and address it.

I'll try and address your points in order:

1. It's hard to say how many locales - this must surely be a function of the target operating system, rather than the host?
2. Having an annotated version seems like a sensible idea - but we'll have to be mindful to only show locales available in the target OS
3. The host locale mapping is fraught. I've come across many a new user PC where they've been using an incorrect but close-enough locale. I suspect a drop-down, while expedient today, might have to be replaced with a map selector a la subiquity.
4. As with (3), automatically using properties from the host gives me cause for concern. I think I'd rather see a deliberate action, where no choice avoids setting any locale directives.

tdewey-rpi:
> Hi :)
> 
> > It's hard to say how many locales - this must surely be a function of the target operating system, rather than the host?
> > That means having a list of supported locales in os_list.json ?
> 
> The problem is the configuration is not os dependent right now. It's global, so if you choose a specific locale that is available on the first os you write, it may not be available in the second one, but you still can press YES when asking for os customization. 
> 
> As the choice YES can be to apply saved customisation for the os you write, having os dependent locales may imply a brand new process for choosing it, hence your reference to subiquity ?

Yes, I'd suggest we might need to work out a schema change for operating systems to advertise their locale options - lest we feed an OS something invalid, and cause a fatal configuration break.

And yes - I'd actually envision this entire mechanism having to change. Locales are actually too important to be relegated to a drop-down, and if you look at the set-up wizards of eg Ubuntu, that's essentially doing what OS customisation is trying to do - and those wizards do it in a much grander way.

For Imager v2.0.0, I'm hoping to introduce something like that - including a full-window Locale selector.

> 
> And if the os get a list of supported locales, maybe it can also declare a list of supported customizations. (an os could support wifi but not username, and so on...

Now that's a good idea (I've been noodling in that direction for a while, and I've broadly come to the conclusion that's the best way forward).

lurch:
I guess this is now defunct in light of Imager 2.0 ?

PR #921, opened 2024-09-11
Update german translation

rejas:


tdewey-rpi (approved):
Danke, @rejas

I note that most of these strings appear to be associated with the Embedded Imager, which is on a different release cadence compared to the Desktop applications - it may take longer for you to see your changes deployed, but they _will_ be deployed!

PR #922, opened 2024-09-12
Correcting a few minor errors in zh-CN (Simplified Chinese) translation

ykla:
Correcting a few minor errors.


tdewey-rpi (approved):
Xie xie, @ykla.

I'm expecting to have to create a v1.9.1 in the near future, so your changes will be included there.

PR #928, opened 2024-09-15
Added Hebrew translation

yarons:
Translated using Lokalize, finalized with Qt Linguist.


tdewey-rpi (approved):
I noticed that 'OS Images' are typically reverse-translated to 'characters' - is this more idiomatic Hebrew?

yarons:
@tdewey-rpi Hebrew only has one letter case, there are no capitals and small/regular letters, there are acronyms and I can theoretically create an acronym from the term מערכת הפעלה which will be מ״ה, the only problem is, most users don't know what it means, in certain circumstances I'm adding the acronym in addition to the full term as part of a longer explanation but I'm this case it's better to just leave it as is, this is the popular form.

PR #929, opened 2024-09-17
Fix GitHub Actions and add Fedora workflow

SingingBush:
While in the process of creating a GitHub workflow for Fedora builds I found that there were problems with the current workflows and the CMakeLists file. I've fixed the problems so that all builds now pass and also, by using a build matrix, the builds are run on multiple versions of the operating systems. They also make build artifacts available in a zip file:

image

Adding Windows builds would be a good addition

### DIFF
--- a/.github/workflows/build-fedora.yml
+++ b/.github/workflows/build-fedora.yml
@@ -1,0 +1,48 @@
+name: Build Fedora
+
+on:
+  push:
+    branches: [ "qml" ]
+  pull_request:
+    branches: [ "qml" ]
+
+
+env:
+  BUILD_TYPE: Release
+
+jobs:
+  build:
+    name: Fedora ${{ matrix.fedora_version }}
+    runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
+      matrix:
+        fedora_version: [ '40', '41', '42' ]
+    container: fedora:${{ matrix.fedora_version }}
+
+    steps:
+    - uses: actions/checkout@v4
+
+    # After Fedora 40 the containers no longer have 'lsblk'
+    - name: Ensure lsblk installed
+      run : sudo dnf install -y util-linux
+
+    - name: Install dependencies
+      # On Fedora, rather than use install-qt-action, install the Qt6 packages from Fedora
+      run: sudo dnf install -y gnutls-devel qt6-qtbase-devel qt6-qtsvg-devel qt6-qtquick3d-devel qt6-qttools-devel qt6-qtquickcontrols2-devel libcurl-devel libarchive-devel libzstd-devel zlib-devel lzma-sdk-devel libdrm-devel
+
+    - name: Configure CMake
+      run: |
+        cmake --version
+        cmake -S src/ -B build/ -Wdev -DCMAKE_BUILD_TYPE=${{env.BUILD_TYPE}}
+
+    - name: CMake Build ${{env.BUILD_TYPE}}
+      run: cmake --build build --target all --config ${{env.BUILD_TYPE}} -- -j4
+
+    - uses: actions/upload-artifact@v4
+      with:
+        name: rpi-imager-fedora-${{ matrix.fedora_version }}.zip
+        if-no-files-found: 'error'
+        path: |
+          build/rpi-imager
+          build/*.qm
--- a/.github/workflows/build-macos.yml
+++ b/.github/workflows/build-macos.yml
@@ -7,28 +7,52 @@ on:
     branches: [ "qml" ]
 
 env:
-  # Customize the CMake build type here (Release, Debug, RelWithDebInfo, etc.)
   BUILD_TYPE: Release
 
 jobs:
   build:
-    runs-on: macos-latest
-
-    defaults:
-      run:
-        working-directory: src
+    # macos-latest is currently macos-14-arm64
+    name: macOS ${{ matrix.os_version }}
+    runs-on: macos-${{ matrix.os_version }}
+    strategy:
+      fail-fast: false
+      matrix:
+        include:
+          - { os_version: '12', arch: 'x86_64' } # macos-12 can only target x86_64
+          - { os_version: '13', arch: 'x86_64' } # macos-13 can do a universal build targeting x86_64 and/or arm64
+          - { os_version: '14', arch: 'arm64' }  # macos-14 is an m1 cpu and can only target arm64
 
     steps:
-    - uses: actions/checkout@v3
+    - uses: actions/checkout@v4
 
     - name: Install Qt
-      uses: jurplel/install-qt-action@v3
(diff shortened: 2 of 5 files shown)


src/CMakeLists.txt
```diff
@@ -465,9 +474,18 @@ elseif(APPLE)
         endif(IMAGER_SIGNING_IDENTITY)
     else()
         # Unsigned application
+        message(STATUS "App will be unsigned")
+        add_custom_command(TARGET ${PROJECT_NAME}
+            POST_BUILD
+            COMMAND "${MACDEPLOYQT_EXECUTABLE}" "${CMAKE_BINARY_DIR}/${PROJECT_NAME}.app" -qmldir="${CMAKE_CURRENT_SOURCE_DIR}" -always-overwrite -no-strip -dmg)
+        # Rename .app file
```
tdewey-rpi, line 481: If you wanted to do this, you probably wanted to remove -dmg above - no point in creating a dmg containing the original name.

src/CMakeLists.txt
```diff
@@ -465,9 +474,18 @@ elseif(APPLE)
...
+        add_custom_command(TARGET ${PROJECT_NAME}
+            POST_BUILD
+            COMMAND mv "${CMAKE_BINARY_DIR}/${PROJECT_NAME}.app" "${CMAKE_BINARY_DIR}/Raspberry\ Pi\ Imager.app")
+        message(STATUS "Creating .dmg file...")
         add_custom_command(TARGET ${PROJECT_NAME}
             POST_BUILD
-            COMMAND "${MACDEPLOYQT_EXECUTABLE}" "${CMAKE_BINARY_DIR}/Raspberry\ Pi\ Imager.app" -qmldir="${CMAKE_CURRENT_SOURCE_DIR}" -always-overwrite -no-strip -dmg)
+            COMMAND hdiutil create -volname "Raspberry Pi Imager" -srcfolder "${CMAKE_BINARY_DIR}/Raspberry\ Pi\ Imager.app" -ov -format UDBZ "${CMAKE_BINARY_DIR}/Raspberry\ Pi\ Imager.dmg")
```
tdewey-rpi, line 488: I'm not entirely sure why you're concerned about the name of the DMGs in the un-notarized case - they're not safe to deploy widely, so their creation seems academic?

src/CMakeLists.txt
```diff
@@ -1,7 +1,10 @@
 # SPDX-License-Identifier: Apache-2.0
 # Copyright (C) 2020-2024 Raspberry Pi Ltd
 
-cmake_minimum_required(VERSION 3.15)
+# To build for Apple platforms you need to set cmake_minimum_required() to 3.21.1 or newer
```
tdewey-rpi, line 4: I've not encountered this, and I'm building for macOS as my default working OS.

.github/workflows/build-ubuntu-deb.yml
```diff
@@ -6,17 +6,62 @@ on:
...
     - name: Install dependencies
-      run: sudo apt install -y --no-install-recommends build-essential devscripts debhelper cmake git libarchive-dev libcurl4-gnutls-dev
-           qtbase5-dev qtbase5-dev-tools qtdeclarative5-dev libqt5svg5-dev qttools5-dev libgnutls28-dev
-           qml-module-qtquick2 qml-module-qtquick-controls2 qml-module-qtquick-layouts qml-module-qtquick-templates2 qml-module-qtquick-window2 qml-module-qtgraphicaleffects
+      run: |
+        sudo apt update -y
+        sudo apt install -y --no-install-recommends devscripts debhelper cmake libarchive-dev libcurl4-gnutls-dev libgnutls28-dev
+
+    - name: Install Qt
+      uses: jurplel/install-qt-action@v4
```
tdewey-rpi, line 30: Using an external action isn't something I'm comfortable with at all.

.github/workflows/build-ubuntu-deb.yml
```diff
@@ -6,17 +6,62 @@ on:
...
+
 jobs:
   build:
-    runs-on: ubuntu-latest
+    name: Ubuntu ${{ matrix.ubuntu_version }}
+    runs-on: ubuntu-${{ matrix.ubuntu_version }}
+    strategy:
+      fail-fast: false
+      matrix:
+        ubuntu_version: [ '20.04', '22.04', 'latest' ]
```
tdewey-rpi, line 19: We no longer support 20.04, nor do many of our dependencies (notably, Qt6)

.github/workflows/build-macos.yml
```diff
@@ -7,28 +7,52 @@ on:
...
+          - { os_version: '14', arch: 'arm64' }  # macos-14 is an m1 cpu and can only target arm64
 
     steps:
-    - uses: actions/checkout@v3
+    - uses: actions/checkout@v4
 
     - name: Install Qt
-      uses: jurplel/install-qt-action@v3
-
+      uses: jurplel/install-qt-action@v4
```
tdewey-rpi, line 29: As below, I'm not comfortable using actions defined outside of raspberrypi.

.github/workflows/build-fedora.yml
```diff
@@ -0,0 +1,48 @@
...
+
+    steps:
+    - uses: actions/checkout@v4
+
+    # After Fedora 40 the containers no longer have 'lsblk'
+    - name: Ensure lsblk installed
+      run : sudo dnf install -y util-linux
+
+    - name: Install dependencies
+      # On Fedora, rather than use install-qt-action, install the Qt6 packages from Fedora
```
tdewey-rpi, line 31: This is not acceptable, and may produce fedora-specific bugs.

.github/workflows/build-fedora.yml
```diff
@@ -0,0 +1,48 @@
...
+  pull_request:
+    branches: [ "qml" ]
+
+
+env:
+  BUILD_TYPE: Release
+
+jobs:
+  build:
+    name: Fedora ${{ matrix.fedora_version }}
```
tdewey-rpi, line 15: Fedora isn't an OS I'm looking to provide explicit support for, so this is probably best suited to a fork.

tdewey-rpi (commented):
PR Rejected.

I wish you'd contacted me before going down this path. There are numerous issues, most of which are a matter of how I intend to develop and release the application rather than technical challenges.

tdewey-rpi:
Expanding on the above comment:

- We don't use github actions today - the simplest path would be to completely remove these files, and I'll be doing so as a follow-up to this PR.
- Fedora is not an explicitly supported platform, and I don't intend to change that at this time. It benefits from the AppImage as used elsewhere.
- The Ubuntu releases are wider than Qt6 supports
- These actions don't produce AppImages
- These changes would introduce a 3P action into our build pipeline - an unacceptable supply chain risk.

PR #931, opened 2024-09-18
Roll-up: v1.9.1

tdewey-rpi:
v1.9.1 introduces support for a debian package wrapper for the AppImage - which requires a manual step to use (renaming the AppImage to `rpi-imager.AppImage`). This mechanism is designed for use in Raspberry Pi OS, allowing the AppImage to be used without additional supporting infrastructure, re-using the existing infrastructure inside Raspberry Pi OS.

This mechanism may be a useful reference to users impacted by 'distribution build breakage', as AppImages still require an update mechanism.

Additionally, some quality of life changes:

1. QML fixes for out-of bounds rendering on Windows
2. QML fixes for mouse behaviours (selecting, cursor choice)
3. Remove extraneous installation directives to further reduce the size of the packages
4. Fixup the help text


src/dependencies/xz-5.6.2/CMakeLists.txt
```diff
@@ -167,6 +167,8 @@ if(NOT MSVC)
     endif()
 endif()
 
+OPTION(XZ_INSTALL OFF "Enable xz Install directives")
```
pastalian, line 170: Does `add_subdirectory(dependencies/xz-5.6.2 EXCLUDE_FROM_ALL)` not work? This way, we can exclude everything. Other dependencies should be able to be excluded in the same way.

Sorry for the late comment.

PR #933, opened 2024-09-19
Add USB Ethernet Gadget toggle

paulober:


tdewey-rpi (changes requested):
Looks great - a handful of cleanup tasks, but otherwise a great addition

src/CMakeLists.txt
```diff
@@ -278,7 +278,8 @@ set(SOURCES ${SOURCES} ${CMAKE_CURRENT_BINARY_DIR}/translations.qrc ${QM_FILES})
 
 if (WIN32)
     # Adding WIN32 prevents a console window being opened on Windows
-    add_executable(${PROJECT_NAME} WIN32 ${SOURCES} ${HEADERS} ${DEPENDENCIES})
+    add_executable(${PROJECT_NAME} WIN32 ${SOURCES} ${HEADERS} ${DEPENDENCIES}
+        extraFiles.qrc)
 else()
     add_executable(${PROJECT_NAME} ${SOURCES} ${HEADERS} ${DEPENDENCIES})
 endif()
```
lurch, line 285: Does the `extraFiles.qrc` also need to be added in the `else` clause here?

lurch:
IIRC this USB Ethernet Gadget mode only works for those Pis where there's not a USB Hub chip in between the SoC and the USB port (e.g. it will work on Pi Zero or Pi 3A+, but won't work on Pi 2B or Pi 3B+). And given that RPi Imager now asks the user which Pi they're going to be running on, I guess we _could_ make this option conditional on that? :thinking:  (or at least make it not-selectable on those Pis where we know it definitely **won't** work?)

src/CMakeLists.txt
tdewey-rpi: Yes, to catch the linux & macOS builds.

src/OptionsPopup.qml
```diff
@@ -826,6 +832,40 @@ Window {
...
+            addCmdline("modules-load=dwc2,g_ether")
+            // TODO: generate mac addresses if there are issues with DHCP
+            //addCmdline("g_ether.dev_addr=8A:89:6a:8d:14:22")
+            //addCmdline("g_ether.host_addr=6A:89:6a:8d:14:22")
+            addCmdline("g_ether.idVendor=0x04b3")
+            addCmdline("g_ether.idProduct=0x4010")
+            addCmdline("g_ether.iManufacturer=\"Raspberry Pi\"")
+            addCmdline("g_ether.bcdDevice=0x0100")
+            addCmdline("g_ether.iProduct=\"USB Ethernet Gadget\"")
+            // TODO: maybe set device serial
```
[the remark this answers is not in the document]
paulober: Right, I now implemented a solution using config files for the deb package. 
Maybe we can also use this approach here and then replace the serial number in the config file with the correct one.
This would require the use to reboot after first boot but it would reduce all these config values from cmdline.txt

src/cli.cpp
```diff
@@ -208,7 +208,7 @@ int Cli::run()
             return 1;
         }
 
-        _imageWriter->setImageCustomization("", "", firstRunScript, "", "");
+        _imageWriter->setImageCustomization("", "", firstRunScript, "", "", false);
```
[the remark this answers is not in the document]
paulober: I already though of this but skipped it for now. I'll have a look.

src/extraFiles/configure-usb-ether-gadget-once.sh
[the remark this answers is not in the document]
paulober: Nope. Removed.

paulober:
Ok, I will implement that.
But concerning 3b+ (@lurch):
- So does it have no OTG support on none of its USB ports? 
- Is there no option to get the USB port into peripheral mode?

src/OptionsPopup.qml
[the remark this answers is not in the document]
paulober: Implemented in last commit. And since my Pi 5 reboots multiple times after inserting a fresh setup this reboot requirement isn't noticeable.

src/OptionsPopup.qml
```diff
@@ -635,6 +637,18 @@ Window {
             }
         }
 
+        var hwFilterList = imageWriter.getHWFilterList()
+        var hwFilterIsModelZero = imageWriter.getHWFilterIsModelZero()
+
+        if (hwFilterList) {
+            var targetTags = ["pi5-64bit", "pi4-64bit", "pi5-32bit", "pi4-32bit"]
```
lurch, line 644: @tdewey-rpi Would it make sense for this to be added as an additional property in the JSON file, rather than having it hardcoded inside Imager itself? (e.g. perhaps a future Raspberry Pi 7 might also support USB-OTG :wink: )

src/imagewriter.h
```diff
@@ -186,6 +195,8 @@ protected slots:
     QJsonDocument _completeOsList;
     QJsonArray _deviceFilter;
     bool _deviceFilterIsInclusive;
+    /* As there is no distinction between normal pi models and zeros (in the tags), this flag can be used to differenciate */
```
lurch, line 201: differenciate -> differentiate

src/downloadthread.cpp
```diff
@@ -974,6 +979,25 @@ bool DownloadThread::_customizeImage()
             _cmdline += " systemd.run=/boot/firstrun.sh systemd.run_success_action=reboot systemd.unit=kernel-command-line.target";
         }
 
+        if (_enableEtherGadget) {
+            // load files from disk and write
+            QByteArray networkConfig = _fileGetContentsTrimmed("://extraFiles/10-usb.network");
+            fat->writeFile("10usb.net", networkConfig);
+            // little optimization for memory constraint systems
```
lurch, line 986: constraint -> constrained ?

src/OptionsPopup.qml
```diff
@@ -826,6 +846,20 @@ Window {
 
             addCmdline("cfg80211.ieee80211_regdom="+fieldWifiCountry.editText)
         }
+        if (chkUSBEther.checked) {
+            // keep parity with cli.cpp
```
lurch, line 850: This "keep parity with..." feels a bit fragile? It's probably something that you should discuss with @tdewey-rpi  , but I wonder if it might be worth extracting out a common function that `OptionsPopup.qml` and `cli.cpp` could both call?

src/extraFiles/g_ether.conf
```diff
@@ -0,0 +1 @@
+options g_ether idVendor=0x04b3 idProduct=0x4010 iManufacturer="Raspberry Pi" bcdDevice=0x0100 iProduct="USB Ethernet Gadget" iSerialNumber=<serial>
```
lurch, line 1: Are these VID and PID values significant, or are they arbitrarily chosen?

src/OptionsPopup.qml
tdewey-rpi: That would be ideal, if we were operating across a single language - however, we are not - and that means any shared class between OptionsPopup and cli.cpp will have to be bound into javascript types and the QML engine - a lot of change for a small duplicated section.

src/extraFiles/g_ether.conf
tdewey-rpi: They are significant - they have to match a known-good Windows-included USB ethernet driver. ISTR, these match to an IBM part.

src/OptionsPopup.qml
lurch: Fair enough.

src/main.qml
```diff
@@ -1614,7 +1614,7 @@ ApplicationWindow {
             }
         }
 
-        imageWriter.setHWFilterList(hwmodel.tags, inclusive)
+        imageWriter.setHWFilterList(hwmodel.tags, inclusive, hwmodel.name.toLowerCase().includes("zero"))
```
tdewey-rpi, line 1617: This makes me nervous, but I can see the motive.

XECDesign:
Shouldn't features like this come after cloud-init support? Why is it enabling a different network configuration service when NM is used for all other interfaces? rpi-imager implementing things that should happen on the OS side instead has only caused headaches in the past.

tdewey-rpi:
> Shouldn't features like this come after cloud-init support? Why is it enabling a different network configuration service when NM is used for all other interfaces? rpi-imager implementing things that should happen on the OS side instead has only caused headaches in the past.

This is a pretty straightforward win that can tide people over until the cloud-init work is completed on the OS (something else @paulober might want to look at).

An earlier version _did_ use a NetworkManager configuration file - @paulober, is there any reason that hasn't made it into this PR?

src/extraFiles/g_ether.conf
paulober: Btw. @tdewey-rpi I tried the ASUS ones but it did not work therefor I fall back to the IBM one.

paulober:
> An earlier version _did_ use a NetworkManager configuration file - @paulober, is there any reason that hasn't made it into this PR?

I have tried creating a NM config file but didn't work so I defaulted back to my systemd-networkd config since it's still on all Rbpi OSs and I haven't found any good resource to get the NM config up and running without wasting to much time. (my approach can be found in the [DEB package repo](https://github.com/paulober/rpi-usb-ether-gadget/blob/bf3e45b53b7501b098e9cf3c5cf83c12ae75803c/manual/usb0.nmconnection))
But I can look into it again, if this is a problem?

src/extraFiles/g_ether.conf
paulober: But I found some thing about a Linux Foundation USB ethernet gadget driver which should be in Windows. But I have to do a bit more research on this.

src/OptionsPopup.qml
paulober: _(Putting it into a resource or adding these lines later in the process before writing the files (if the bool is true) may be a possible option...)_

src/downloadthread.cpp
paulober: thanks will update this

src/imagewriter.h
paulober: thanks will update this

src/OptionsPopup.qml
paulober: Here an example of some devices using the new scheme:
```json
{

"devices": [
      {
        "name": "Raspberry Pi 5",
        "tags": [
          "pi5-64bit",
          "pi5-32bit"
        ],
        "icon": "https://downloads.raspberrypi.com/imager/icons/RPi_5.png",
        "description": "The latest Raspberry Pi, Raspberry Pi 5",
        "matching_type": "exclusive",
        "features": [
            "ether_gadget"
        ]
      },
      {
        "name": "Raspberry Pi 4",
        "tags": [
          "pi4-64bit",
          "pi4-32bit"
        ],
        "default": false,
        "icon": "https://downloads.raspberrypi.com/imager/icons/RPi_4.png",
        "description": "Models B, 400, and Compute Modules 4, 4S",
        "matching_type": "inclusive",
        "features": [
            "ether_gadget"
        ]
      }
]

}
```

paulober:
The current setup has one drawback: On the RPi zero if this config is applied the USB port can only be used of this ether gadget and not for HID devices like keyboard and mouse. This would require the manual config approach with `libcomposite` as the current gadgets in the kernel only support either ethernet or HID or ethernet + serial (there are also more but out of the topic)

src/OptionsPopup.qml
tdewey-rpi: I like it. We can also then use those feature flags in an AND with feature flags exposed by the OS to identify if a feature should be supported.

lurch:
If people are making custom HID devices, perhaps they're more likely to be doing that with Raspberry Pi Pico (microcontroller) rather than Raspberry Pi Zero (computer)? :man_shrugging:

paulober:
> If people are making custom HID devices, perhaps they're more likely to be doing that with Raspberry Pi Pico (microcontroller) rather than Raspberry Pi Zero (computer)? 🤷‍♂️

I'm not sure what you mean by this?

The HID functionality is also need so that normal keyboards and mice work with that port so after enabling this feature they won't be able to connect any of those devices any more until they uninstall the package.

src/OptionsPopup.qml
paulober: What about renaming this to `supportedFeatures` so it doesn't implicate that this OS or this device has this feature?

lurch:
Ahhh, I misunderstood what you were saying - I thought you meant that you wanted to simultaneously enable both USB Ethernet and HID gadget modes, so that the you could SSH from your laptop to your Pi Zero, whilst _also_ allowing the Pi Zero to act as a virtual keyboard input-device to your laptop.

But you're actually saying that if you enable this Ethernet gadget mode (which puts the USB OTG port into device mode), then you can't then later connect a real USB keyboard (which would require the USB OTG port to be in host mode). Isn't that expected behaviour? :thinking:

src/main.qml
paulober: You can rest now sure,. that this check isn't in-use anymore. Has been replaced by new HW Supported Features list.

src/OptionsPopup.qml
lurch: `capabilities` ?

paulober:
@XECDesign
May I ask how 
> rpi-imager implementing things that should happen on the OS side instead has only caused headaches in the past.

is ment?

Would you suggest writing a new kernel USB gadget, adding an option to `raspi-config ` and make `rpi-imager` then just calling `raspi-config` on first boot to enable this feature?

paulober:
> But you're actually saying that if you enable this Ethernet gadget mode (which puts the USB OTG port into device mode), then you can't then later connect a real USB keyboard (which would require the USB OTG port to be in host mode). Isn't that expected behaviour? 🤔

I guess you're right. That sure be the reason why the `g_multi` gadget supports only ethernet, serial and stuff like that. Because they all require the port beeing in the same mode and g_hid would require the port to be in a different mode....

Sorry for the confusion. I guess this is something that needs to be noted somewhere like in the UI and the DEB pkg description + README?

src/OptionsPopup.qml
lurch: Also, given that this is describing a feature of the hardware, perhaps it should be named `"usb_otg"` rather than `"ether_gadget"` ?

lurch:
> g_hid would require the port to be in a different mode

I assume the `g` in `g_hid` still stands for "gadget", i.e. `g_hid` _does_ still run in device-mode rather than host-mode, but it's used for making the Pi Zero appear to your laptop as a virtual keyboard or mouse (etc.) input-device. (Ah yes: https://docs.kernel.org/usb/gadget_hid.html )

XECDesign:
> @XECDesign May I ask how
> 
> > rpi-imager implementing things that should happen on the OS side instead has only caused headaches in the past.
> 
> is ment?
> 

When image customisation was added, it was just the firstrun.sh script doing all the work. This caused problems because it made a ton of assumptions about the OS. When we moved away from dhcpcd, for example, the wifi config option broke.

So I added `/usr/lib/raspberrypi-sys-mods/imager_custom` as something rpi-imager can call instead. Then it doesn't matter how the images changes, as long as `imager_custom` does the right thing.

I have previously proposed that the json file could have a field matching imager_custom version to determine what customisations the image supports. However, since firstrun.sh was just temporary and we should be using cloud-init to do all of this anyway, it was deemed necessary.

But since this isn't a feature I'm comfortable adding to Raspberry Pi OS as-is, it's probably better for rpi-imager to handle it itself in this case.

lurch:
@paulober Something that occurred to me this morning (following on from the earlier HID discussion) - maybe there should also be some shell scripts on the Pi that you can use to easily "enable / disable" this USB Ethernet functionality, so that if you're using USB Ethernet gadget on a Zero and you later want to use a **real** USB keyboard with the same Zero, you could connect to the Zero over SSH over the USB-ethernet, run the "disable usb ethernet" script, and then the next time you boot up the Zero it'll be running in "normal" mode (so you could directly-connect keyboard/mouse/monitor). And if you then later wanted to switch it back into USB Ethernet gadget mode, you could run the "enable usb ethernet" script?

paulober:
@lurch Sounds like a good addition. I'll have a look

src/OptionsPopup.qml
paulober: Completely agree features and also supported features seems a bit confusing. 
And `usb_otg` would also have the advantage that it can also be used for other features in the future to detect if a device OS combi does have support for it rather than checking if it supports `ether_gadget`.

paulober:
The script i just added is not completely ready yet.

src/extraFiles.qrc
```diff
@@ -3,5 +3,6 @@
         <file>extraFiles/10-usb.network</file>
         <file>extraFiles/g_ether.conf</file>
         <file>extraFiles/usb-ether-gadget.conf</file>
+        <file>extraFiles/rpi-usb-ether-gadget.sh</file>
```
lurch, line 6: Minor nitpick, but if this is being written into the image as `/usr/bin/rpi-usb-ether-gadget` then perhaps this filename should be `extraFiles/rpi-usb-ether-gadget` ? (i.e. drop the `.sh` for consistency?)

src/extraFiles/rpi-usb-ether-gadget.sh
```diff
@@ -0,0 +1,50 @@
...
+    exit
+else # add else if to make toggling possible without toggle keyword
+    echo "Usage: rpi-usb-ether-gadget [on|off|toggle|status|help]"
+    exit
+fi
+
+#rm /etc/modprobe.d/g_ether.conf
+
+# check if /etc/modules-load.d/usb-ether-gadget.conf exists to know if to turn of or on or the force flags are set
+if ([ -f /etc/modules-load.d/usb-ether-gadget.conf ] && [ "$FORCE_ON" = false ]) || [ "$FORCE_OFF" = true ]; then
```
lurch, line 40: IMHO this `if` is hard to read, and it'd be better to handle this "toggling logic" as part of the argument-parsing done earlier? So that this just becomes `if [ "$TURN_ON" = true ]; then`

src/extraFiles/rpi-usb-ether-gadget.sh
```diff
@@ -0,0 +1,50 @@
...
+fi
+
+# check if the script is run on a Raspberry Pi
+if ! grep -q "Raspberry Pi" /proc/device-tree/model; then
+    echo -e "\e[31mThis script is for Raspberry Pi only\e[0m"
+    exit
+fi
+
+FORCE_ON=false
+FORCE_OFF=false
```
lurch, line 16: Might be more sensible to have a single `TURN_ON` variable, rather than separate `FORCE_ON` and `FORCE_OFF` variables?

src/extraFiles/rpi-usb-ether-gadget.sh
```diff
@@ -0,0 +1,50 @@
...
+    fi
+    exit
+else # add else if to make toggling possible without toggle keyword
+    echo "Usage: rpi-usb-ether-gadget [on|off|toggle|status|help]"
+    exit
+fi
+
+#rm /etc/modprobe.d/g_ether.conf
+
+# check if /etc/modules-load.d/usb-ether-gadget.conf exists to know if to turn of or on or the force flags are set
```
lurch, line 39: "turn of" -> "turn off" :wink:

src/extraFiles.qrc
paulober: I added it to the resource with .sh for better detection in editors.

src/extraFiles/rpi-usb-ether-gadget.sh
paulober: fixed

src/imagewriter.cpp
```diff
@@ -502,16 +502,16 @@ bool ImageWriter::getHWFilterListInclusive() {
     return _deviceFilterIsInclusive;
 }
 
-bool ImageWriter::andSupportedFeatures(const QString &feature) {
-    return this->checkHWFeatureSupport(feature) && this->checkSWFeatureSupport(feature);
+bool ImageWriter::andCapabilities(const QString &cap) {
+    return this->checkHWCapability(cap) && this->checkSWCapability(cap);
```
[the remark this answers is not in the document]
paulober: 1. Fixed
2. I now added a second optional parameter that would allow to specify different naming and still has the option to only specifying a unified naming.

macmpi:
How about replacing `g_ether` by `g_cdc` in `rpi-usb-ether-gadget.sh` & `usb-ether-gadget.conf`: this would give both ethernet and serial options, each having own merits to setup PiZero & al devices?
(note; `dwc2` in `usb-ether-gadget.conf` is probably not required as this driver is loaded by setting dtoverlay)

leongross:
Is this still in active development? I would really like to see this feature come in.

tdewey-rpi:
Dropping, as this PR has been superseded by implemented changes on the 2.0 branch.

PR #948, opened 2024-10-14
Added Portuguese translation

hugok79:
* Add Portuguese translation
* Sort `translations.qrc` file alphabetically


tdewey-rpi (approved):
Verified with machine translation all translated strings.

Thanks for being so thorough!

Issue #746, opened 2023-11-30
[BUG] New Icon Doesn't Match Apple Design Guidelines


Logicer16:
**Describe the bug**
Following up on #695, the new icon is too large and has a smaller corner radius than what [Apple's Human Interface Guidelines](https://developer.apple.com/design/human-interface-guidelines/app-icons#macOS). This is contrary to the goal stated in https://github.com/raspberrypi/rpi-imager/issues/695#issuecomment-1773740128:
> the desire was [an] icon ... that also complied with the Apple HI guidance.

**To Reproduce**
Include steps to allow us to reproduce the error. Include your host OS, and the version of Raspberry Pi Imager.
1. View the imager's icon next to icons of other Mac apps.

**Expected behaviour**
The new icon to meet the goal of conforming to the Apple Human Interface Guidelines.

**Screenshots or video**
Comparing it with other icons in the dock makes the discrepancy quite striking
<img width="540" alt="Screen Shot 2023-11-30 at 5 56 44 pm" src="https://github.com/raspberrypi/rpi-imager/assets/81077770/f8784b18-4df0-49a0-ac07-06338b00b1c4">
Placing the icon against Apple's provided template shows exactly by how much the new icon is off.
<img width="811" alt="Screen Shot 2023-11-30 at 6 09 43 pm" src="https://github.com/raspberrypi/rpi-imager/assets/81077770/97f45dfb-3e61-4854-8c5c-ebe6c6b83a0e">
The micro SD card is fine as is though as it is a tool external from the main icon, which is permitted by the guidelines.

**Desktop (please complete the following information):**
 - OS: macOS 12.7.2

**The name of the OS you are trying to write**
N/A

**Are you using OS Customisation?**
N/A

**Additional context**
N/A

tdewey-rpi:
Thanks for the report, @Logicer16.

@jackbenwillis Have I mangled your asset in the merge?

tdewey-rpi:
Also, thank you @Logicer16 for taking the time to assemble a visual aide. This is greatly appreciated when reporting UI concerns in particular.

tdewey-rpi:
Thanks for the catch, @Logicer16. We'll look to revise the asset, but this is unlikely to happen in 2023.

elsiehupp:
I [mentioned](https://github.com/raspberrypi/rpi-imager/issues/397#issuecomment-1584626517) this on #397:

[...]

PR #949, opened 2024-10-15
Make macOS icon follow Apple's Design Guidelines

Ferbez:
As stated in #746, current macOS icon doesn't follow Apple's Design Guidelines. This icon is made using Apple's own [Design Resources](https://developer.apple.com/design/resources/) - all while using current icon's design.
image


tdewey-rpi:
Unfortunately, I cannot accept a visual asset not created by our communications team.

Paging @jackbenwillis for visibility.

jackbenwillis:
We hope to get onto this soon. It's on our radar but pretty low on our list of priorities

PR #959, opened 2024-11-08
OS Customisation: Regex to match ssh public keys

### LINKED ISSUE #956: [BUG]: firstrun.sh: line72: unexpected eof while looking for matching `''
### What happened?

Hello,
I flashed raspberry os lite (I tried 64 and 32bits) on a raspberry pi 3 with imager v1.8.5 and with personnal settings like ssh key, wifi credentials etc ...
My raspberry stop at boot after lanching firstrun.sh, this is the error I see on my screen:

> /boot/firmware/firstrun.sh: line72: unexpected eof while looking for matching `''

after that line kernel-command-line.service failed to start and the os is powered off with shutdown.target

I also tried to flash raspbian with my settings and it's working

### Version

1.8.5

macOS

### Host OS Version

Sequoia, 15.1

### Selected OS

Raspberry Pi OS Lite Bookworm 64 bits

Raspberry Pi 3B, 3A+, 3B+, and Compute Module 3, 3+

### OS Customisation

### Relevant log output

```shell
/boot/firmware/firstrun.sh: line72: unexpected eof while looking for matching `''
```

tdewey-rpi:
Fixes #956

Previously we would just trust that users would include a correct SSH public key. Unfortunately, this isn't a viable stategy.

So, let's introduce a regex to perform some light validation of the ssh-key, to avoid having boot failures.


src/OptionsPopup.qml
```diff
@@ -404,15 +404,14 @@ Window {
...
+                TextField {
                     id: fieldPublicKey
                     enabled: radioPubKeyAuthentication.checked
-                    textFormat: TextEdit.PlainText
                     wrapMode: TextEdit.WrapAnywhere
                     Layout.fillWidth: true
                     Layout.minimumWidth: 350
                     Layout.leftMargin: 40
-                    selectByMouse: true
+                    validator: RegularExpressionValidator { regularExpression: /^ssh-(ed25519|rsa|dss|ecdsa) AAAA(?:[A-Za-z0-9+\/]{4})*(?:[A-Za-z0-9+\/]{2}==|[A-Za-z0-9+\/]{3}=|[A-Za-z0-9+\/]{4})( [^@]+@[^@]+)?/ }
```
lurch, line 414: If the aim of this PR is to fix #956 by preventing `'` or `"` from  appearing in the public key field (which breaks `firstrun.sh`), perhaps the use of `[^@]+` is a bit overly broad? :thinking:

src/OptionsPopup.qml
```diff
@@ -404,15 +404,14 @@ Window {
                     // textFormat: Text.PlainText
                     Layout.leftMargin: 40
                 }
-                TextArea {
+                TextField {
```
lurch, line 407: Does the switch from `TextArea` to `TextField` mean that advanced users will no longer be able to specify _multiple_ public SSH keys in this field?

tdewey-rpi: Yes - this breaks that case in favour of introducing a real-time validator. TextArea has no such equivalent mechanism.

maxnet: Do keep in mind that support for multiple keys was an explicit user request, so this is likely to lead to complaints. It was once a textfield and people opened issues about that: https://github.com/raspberrypi/rpi-imager/issues/316

src/OptionsPopup.qml
tdewey-rpi: > If the aim of this PR is to fix #956 by preventing `'` or `"` from appearing in the public key field (which breaks `firstrun.sh`), perhaps the use of `[^@]+` is a bit overly broad? 🤔

You're correct - this would prevent one appearing at the _start_ of the ssh-key, but not at the end.

tdewey-rpi:
Screenshot 2024-11-18 at 15 08 39

The new PR includes a dynamic UI, where each key gets added on it's own line, with it's own validator. It includes a few useful side-effects:

- Each tab gets it's own ScrollView, so position is not conserved across tabs - handy for multiple keys
- Keys can be trivially deleted alone - so no more hunting through a TextArea to find the *one key* you want to remove
- Validation is performed live - so entering an invalid key is not possible.

The change isn't as graceful as I would have liked - but it does appear to do the right thing, and may be revisited later.

lurch:
Oooh, snazzy :star2:  I assume that even with the option to add multiple keys, it'll only let you click "RUN SSH-KEYGEN" once? If you've already added a couple of keys by hand, and then click "RUN SSH-KEYGEN", will it add the newly-generated key to the bottom of the list?
I wonder if it'd make sense for the "ADD SSH KEY" button to be disabled if the textbox for the last key in the list is empty?

src/OptionsPopup.qml
```diff
@@ -538,10 +628,34 @@ Window {
             }
         }
         if ('sshAuthorizedKeys' in settings) {
-            fieldPublicKey.text = settings.sshAuthorizedKeys
+            console.log("sshAuthorizedKeys: " +JSON.stringify(settings.sshAuthorizedKeys))
```
lurch, line 631: Hopefully all these `console.log`s will get removed before this PR gets merged, otherwise we might upset the tinfoil-hat brigade :wink:

lurch:
> multi-paste

Oh, so if I click on the "Add key" button 4 times, and then paste 4 lines of text, those 4 lines will automatically populate one text-box each?

tdewey-rpi:
> > multi-paste
> 
> Oh, so if I click on the "Add key" button 4 times, and then paste 4 lines of text, those 4 lines will automatically populate one text-box each?

No - because in that case there's no good way to feedback that one of the keys early in the chain has failed. I'm imagining more a situation where:

(I have 5 keys to add) -> (I click "Add Key" 5 times) -> for n..5 (I copy-paste a key from my key storage into an empty field, and if it fails to paste, I note the error and correct my behaviour - or discard that key)

lurch:
Right, I misunderstood what you meant by "multi-paste", sorry.

PR #964, opened 2024-11-21
deps: cURL 8.{8->11}.0

tdewey-rpi:
Security update. https://curl.se/docs/vuln-8.8.0.html


tdewey-rpi:
Confirmed this builds and executes correctly on macOS. Merging ahead of validation on Windows and Linux.

PR #969, opened 2024-12-02
Filter system drives by user choice

tdewey-rpi:
Rather than using a compile-time variable to determine when to filter system drives, introduce a UI item to allow users to dynamically grey-enable system drives.

This change carries the risk that users might decide to destroy their PC's hard drive. The mitigation has been to retain the behaviour of filtering by default - this change only moves the point of choice on filtering to the UI.

This change will be most beneficial in the Network Installer, where attached NVMe or USB drives may incorrectly be flagged as 'system' drives.

### DIFF
--- a/src/main.qml
+++ b/src/main.qml
@@ -1099,7 +1099,6 @@ ApplicationWindow {
                             Layout.fillWidth: true
                             font.family: roboto.name
                             font.pointSize: 16
-                            color: (isReadOnly || unselectable) ? "grey" : "";
                             color: !dstitem.unselectable ? "" : "grey";
                             text: {
                                 var sizeStr = (size/1000000000).toFixed(1)+ " " + qsTr("GB");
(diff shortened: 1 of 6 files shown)


src/main.qml
```diff
@@ -1095,11 +1113,13 @@ ApplicationWindow {
...
                             font.pointSize: 12
-                            color: "grey"
+                            color: !dstitem.unselectable ? "" : "grey";
                             text: {
                                 var txt= qsTr("Mounted as %1").arg(mountpoints.join(", "));
                                 if (isReadOnly) {
-                                    txt += " " + qsTr("[WRITE PROTECTED]")
+                                    txt += " " + qsTr("[WRITE PROTECTED]");
+                                } else if (isSystem) {
+                                    text += " [" + qsTr("SYSTEM") + "]";
```
lurch, line 1122: Interesting that `" " +qsTr("[WRITE PROTECTED]")` has the `[ ]` _inside_ the translatable string, but " [" + qsTr("SYSTEM") + "]" has the `[ ]` outside of the translatable string.
(I guess the latter is preferable, but we can't change the former to match without "breaking" a bunch of already-existing translations?)

lurch:
What's the "lifetime" of the "Exclude System Drives" remaining unticked? Does it become "re-ticked" every time the user closes and then opens the drive-selection dialog? (which is probably the "safest" option).
If the user selects a "system drive" and then re-ticks the "Exclude System Drives" button, does the selected system drive automatically get deselected?

PR #970, opened 2024-12-02
Make 'gigabytes' translatable

tdewey-rpi:
An egregious error in the description field of the storage selection items meant that regardless of your chosen language, you got 'gigabytes' in English.

In the spirit of international cooperation, make this string translatable too.


lurch:
Do all languages always have "gigabytes" _after_ the number? Are there any languages which display the (translated version of) gigabytes _before_ the number itself?
See https://doc.qt.io/qt-5/i18n-source-translation.html#using-numbered-arguments and/or https://doc.qt.io/qt-5/i18n-source-translation.html#handling-plurals

tdewey-rpi:
Not all, but I’m only aware of French carrying a different extension.
There’s additional work to enable that language to also be correct here -
they use Giga Octets (pow2), rather than the current calculation.

PR #975, opened 2024-12-06
AppImage: Disable Wayland support

### LINKED ISSUE #942: [BUG]: symbol lookup error with 1.9.0 and mesa 24.2.2-1 ( Kali Linux host, Debian testing equivalent)
### What happened?

When attempting to run the aarch64 1.9.0 AppImage, I am getting an unknown symbol

```
steev@finn:~/Downloads$ ./Raspberry_Pi_Imager-1.9.0-aarch64.AppImage
/tmp/.mount_RaspbeVM7Y30/AppRun.wrapped: symbol lookup error: /lib/aarch64-linux-gnu/libEGL_mesa.so.0: undefined symbol: wl_display_create_queue_with_name
```

I am guessing this went away in newer mesa, the version of libegl-mesa0 is 24.2.2-1 from Debian testing.

### Version

1.9.0 (Default)

Windows

### Host OS Version

Kali Linux 

### Selected OS

Kali Linux

Raspberry Pi 5

### OS Customisation

### Relevant log output

_No response_

tdewey-rpi:
The Wayland composition flags in the Qt AppImage support are not safe to use - they include a system-owned copy of libwayland-client that is cross-called by the system Mesa - outside of the AppImage. Including the system-owned copy from the build host creates an ABI breakage that prevents execution.

This is an unacceptable situation, and until a more precise fix can be provided, disable Wayland client support in Imager.

Fixes #942

### DIFF
--- a/appimagecraft.yml
+++ b/appimagecraft.yml
@@ -20,5 +20,12 @@ appimage:
     raw_environment:
       QML_SOURCES_PATHS: "\"$PROJECT_ROOT\"/src/qmlcomponents/"
     environment:
-      EXTRA_PLATFORM_PLUGINS: "libqwayland-egl.so;libqwayland-generic.so"
-      EXTRA_QT_PLUGINS: "waylandcompositor"
+      # Disable Wayland back-end until the dependency problem gets resolved.
+      # Essentially, importing libqwayland-generic causes the AppImage dependency walker
+      # to find libwayland-client from the build OS, and include that in the AppImage.
+      # This is deleterious - because the the runtime OS graphics stack also depends on that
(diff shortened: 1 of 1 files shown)


lurch:
Does this mean that Raspberry Pi Imager will always be run under XWayland instead?

tdewey-rpi:
On Wayland platforms, yes. The only graphics backend that is safe to ship appears to be the XCB backend - at least until I can persuade the AppImage tooling to reject packaging the Wayland client libraries from the build machine and just package the Qt parts. Even then, with apparently poor ABI guarantees, I'm not convinced that will be safe either.

PR #990, opened 2025-01-08
Fixup no-PSK network customisation.

### LINKED ISSUE #977: [BUG]: Imager 1.9.0 - can't set login password using macOS 15.0.1 or Windows 11
### What happened?

I saw this behavior with Imager 1.9.0 running both on macOS 15.0.1 and on Windows 11. 

I downloaded the imager from this github repo and installed it in the normal way. I then ran the imager, selecting:
- RPi4
- Ubuntu Server 24.04.1 LTS
- The microsd card I had just inserted
- Clicked NEXT and "Edit settings"

I set the following:
- RPi4-at-LLL.local
- ubuntu/ubuntu for credentials (typed extremely carefully, verifying each character)
- Wireless LAN: "HBTL" with no password
- Wireless LAN Country: US
- Time Zone: New York
- Keyboard: US
then clicked Save.

I clicked YES for apply customisation settings, then cicked YES to erasing the media. I authorized the Mac to write the microSD, then it wrote and verified the media in the expected way.

**Problems:**

1. After checking the User/password checkbox, I could not de-select it.

2. More importantly, although the microSD card caused the RPi4 to boot as expected, after being presented with the `login:` prompt, those credentials (ubuntu/ubuntu) did NOT work. I cannot log in.

What other troubleshooting information could I provide? Thanks

### Version

1.9.0

Windows

### Host OS Version

Windows 11; macOS 15.0.1

### Selected OS

Ubuntu server 24.04.1 LTS

microSD Card in a USB reader

### OS Customisation

### Relevant log output

```shell
How would I get log files?
```

tdewey-rpi:
Rather than always serialising the psk fields, conditionally serialise them if the length of the password field text is greater than 0.

Fixes #977 in combination with users waiting for cloud-init to complete.

### DIFF
--- a/src/OptionsPopup.qml
+++ b/src/OptionsPopup.qml
@@ -908,7 +908,11 @@ Window {
                 fieldWifiPassword.text.length < 64
             var cryptedPsk = isPassphrase ? imageWriter.pbkdf2(fieldWifiPassword.text, fieldWifiSSID.text)
                                           : fieldWifiPassword.text
-            wpaconfig += "\tpsk="+cryptedPsk+"\n"
+            if (cryptedPsk.length > 0) {
+                wpaconfig += "\tpsk="+cryptedPsk+"\n"
+            } else {
+                wpaconfig += "\tkey_mgmt=NONE\n"
+            }
             wpaconfig += "}\n"
 
             addFirstRun("if [ -f /usr/lib/raspberrypi-sys-mods/imager_custom ]; then")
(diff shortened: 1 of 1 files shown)


tdewey-rpi:
Dropping this. I can't see that it's going to do anything useful for the cloud_init case - even if it's slightly clearer on intent.

It may help the wpa_supplicant case, however, where key_mgmt=NONE appears to be required for an 'open' network - but we've had no bug in that area.

PR #1024, opened 2025-03-21
make NSIS(windows installer) use the "all users" shell folder - fix app shortcut for all users

marlop352:
- make NSIS(windows installer) use the "all users" shell folder
  - that makes the shortcut be created for all users instead of only the admin user that installed the application


tdewey-rpi (approved):
Thanks for the submission, @marlop352

PR #1035, opened 2025-04-26
fix typo in debug output for firstrun.sh

Secretowo:
The debug output refers to it as "firstuse.sh" instead of "firstrun.sh" like everywhere else. I could not find any other references to this string, so I am assuming its a mistake.


tdewey-rpi:
Thanks for the submission, @Secretowo

PR #1038, opened 2025-04-30
Italian language update

bovirus:
@tdewey-rpi 

Italian language update.
Please merge. Thanks.

### DIFF
--- a/src/i18n/rpi-imager_it.ts
+++ b/src/i18n/rpi-imager_it.ts
@@ -216,48 +216,48 @@ Aggiungi sia &apos;rpi-imager.exe&apos; che &apos;fat32format.exe&apos; all&apos
     <message>
         <location filename="../DstPopup.qml" line="24"/>
         <source>Storage</source>
-        <translation type="unfinished">Scheda SD</translation>
+        <translation>Storage</translation>
     </message>
     <message>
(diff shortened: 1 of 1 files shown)


src/i18n/rpi-imager_it.ts
```diff
@@ -216,64 +216,64 @@ Aggiungi sia &apos;rpi-imager.exe&apos; che &apos;fat32format.exe&apos; all&apos
...
         <location filename="../DstPopup.qml" line="152"/>
         <source>Mounted as %1</source>
-        <translation type="unfinished">Montato come %1</translation>
+        <translation>Montato come %1</translation>
     </message>
     <message>
         <location filename="../DstPopup.qml" line="139"/>
         <source>GB</source>
-        <translation type="unfinished"></translation>
+        <translation></translation>
```
lurch, line 245: Still an empty translation-string here? :confused:

bovirus: Sorry but in my fiule is

Quite strange. 
QTlinguist report as translated.
I will fix this.,

lurch: The ` type="unfinished"` doesn't matter - it's in red, which means it's the old version of the line. It's only the `<translation></translation>` in the green (new) version of the line that I was worried about.

PR #1056, opened 2025-05-25
Add "rpi-sdimg" file extension

flederwiesel:
… to `openFileDialog()` and `getUsbSourceOSlist()` filter.

This is the file extension used by Yocto Project's `sdcard_image-rpi.bbclass`.

### DIFF
--- a/src/imagewriter.cpp
+++ b/src/imagewriter.cpp
@@ -792,7 +792,7 @@ void ImageWriter::openFileDialog()
 
     QFileDialog *fd = new QFileDialog(nullptr, tr("Select image"),
                                       path,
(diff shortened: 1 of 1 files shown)


tdewey-rpi:
Thanks for the submission, @flederwiesel 

Unfortunately I'm not inclined to accept this.

I've reviewed the class, and I can see a much simpler change - just have the Yocto bbclass produce a .img, as every other tool would do. It's quite bizarre that the bbclass doesn't allow overriding that parameter, and it would be a single-character change to upstream with the ability for users to set the filename to whatever they choose.

https://github.com/agherzan/meta-raspberrypi/blob/49a2d124890b96d28a590e52e6d0dd611c114f6b/classes/sdcard_image-rpi.bbclass#L62

PR #1072, opened 2025-06-07
Update French translation

martignoni:


martignoni:
> As a heads-up, there's going to be a range of new strings in 1.9.5, and if you'd like to get an update in ahead of that release, let me know and I'll make sure you're notified.

@tdewey-rpi I'm indeed interested. Please keep me in the loop.

PR #1075, opened 2025-06-09
Fix numerous file leaks, bad queing behaviours.

tdewey-rpi:
- Missing libarchive close calls
- Missing filesystem sync operations
- libarchive handle leaks in exceptional paths
- LocalFileExtractThread didn't close the input file
- DownloadThread cachefile handles leaking
- ImageWriter destructor failed to clean up _thread

### DIFF
--- a/src/imagewriter.cpp
+++ b/src/imagewriter.cpp
@@ -237,6 +237,21 @@ ImageWriter::ImageWriter(QObject *parent)
 
 ImageWriter::~ImageWriter()
 {
+    // Ensure any running thread is properly cleaned up
+    if (_thread) {
+        if (_thread->isRunning()) {
+            qDebug() << "Cancelling running thread in ImageWriter destructor";
+            _thread->cancelDownload();
+            if (!_thread->wait(10000)) {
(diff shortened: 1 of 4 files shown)


src/imagewriter.cpp
maxnet, line 1: Note that if a parent-child relationship is estabilished when creating a QObject, which is the case with _thread (parent = this):
  
```
        _thread = new DownloadExtractThread(urlstr, _dst.toLatin1(), _expectedHash, this);
```

then when the parent object (ImageWriter in this case) is destructed, the destructors of all children are called automatically first in QObject::~QObject().
Note that DownloadThread::~DownloadThread() already has logic to cancel and wait().
Although it currently does not have a timeout there (so do may indeed give problems if for some reason it hangs)

QFile also closes the file in its destructor.
If file still stays open in practice that can be a sign it's parent object is not being destructed properly or it is hanging in a destructor somewhere.

tdewey-rpi:
Holding for test on Windows.

tdewey-rpi:
Dropping in favour of the 1.9.5 roll-up. Testing on Windows confirms this change resolves the issue.

PR #1084, opened 2025-06-16
1.9.5 Fix roll-up

tdewey-rpi:
* Add architecture sorting to allow Raspberry Pi Zero 2 W to get armhf builds recommended, even though it's 64 bit capable.
* Add a new cache management structure to better control the OS image cache - including verification of the cache file before writing.
* Attempt to fix OS customisation horrors by ensuring we use the trimmed values from customisation.
* Resolve a range of file locking problems on Windows
* Fix a translation bug, where after selecting an OS or destination storage device the labels would lose their anchor, and fail to respond to translation change events
* macOS: Programmatically styled DMGs.

### DIFF
--- a/src/windows/acceleratedcryptographichash_cng.cpp
+++ b/src/windows/acceleratedcryptographichash_cng.cpp
@@ -99,7 +99,7 @@ struct AcceleratedCryptographicHash::impl {
         cleanup();
     }
 
-    void cleanup() {
+    void cleanup() const {
         if(hAlg)
         {
             BCryptCloseAlgorithmProvider(hAlg,0);
@@ -162,7 +162,7 @@ struct AcceleratedCryptographicHash::impl {
 private:
     BCRYPT_ALG_HANDLE       hAlg            = NULL;
     BCRYPT_HASH_HANDLE      hHash           = NULL;
-    NTSTATUS                status          = STATUS_UNSUCCESSFUL;
+    mutable NTSTATUS        status          = STATUS_UNSUCCESSFUL;
     DWORD                   cbData          = 0,
                             cbHash          = 0,
                             cbHashObject    = 0;


src/cachemanager.cpp
```diff
@@ -0,0 +1,284 @@
...
+    }
+}
+
+void CacheManager::onDiskSpaceCheckComplete(qint64 availableBytes, const QString& directory)
+{
+    qDebug() << "Disk space check complete:" << availableBytes / (1024*1024*1024) << "GB available in" << directory;
+    
+    updateCacheStatus([&](CacheStatus& status) {
+        status.availableBytes = availableBytes;
+        status.hasAvailableSpace = availableBytes > (1024 * 1024 * 1024); // > 1GB
```
lurch, line 135: Should this "1GB minimum free space" value be stored in `config.h`?

src/cachemanager.cpp
```diff
@@ -0,0 +1,284 @@
...
+    
+    emit verificationComplete(isValid, fileName, expectedHash);
+}
+
+void CacheVerificationWorker::checkDiskSpace()
+{
+    qDebug() << "Background: Checking disk space for cache operations";
+    
+    // Ensure cache directory exists
+    if (!ensureCacheDirectoryExists()) {
```
lurch, line 233: Perhaps a micro-optimisation might be to record whether the cache directory existed before this call, and then delete the newly-created cache directory if there's less than 1GB free space and the cache directory didn't previously exist? (but of course the overhead of adding that code probably isn't worth it)

doc/json-schema/os-list-schema.json
```diff
@@ -137,6 +137,102 @@
...
+                    "items": {
+                        "$id": "#/properties/imager/properties/devices/items",
+                        "type": "object",
+                        "title": "Device entry",
+                        "description": "An entry for a Raspberry Pi device.",
+                        "default": {},
+                        "examples": [
+                            {
+                                "name": "Raspberry Pi 4",
+                                "description": "4GB, 8GB and 2GB variants",
```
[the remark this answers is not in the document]
tdewey-rpi: This is an example of a field that will likely never be set by a user - the precise contents don't really matter.

src/downloadextractthread.cpp
```diff
@@ -145,17 +100,48 @@ DownloadExtractThread::~DownloadExtractThread()
...
+    static QElapsedTimer timer;
+    bool firstProgressUpdate = false;
+    if (!_progressStarted) {
+        _progressStarted = true;
+        firstProgressUpdate = true;
+        timer.start();
+        qDebug() << "Started progress updates after successful drive opening";
+    }
+    
+    // Only emit progress updates every 100ms to avoid flooding (but always emit the first one)
```
lurch, line 114: "at most every 100ms" ? (i.e. the updates may be less frequent for a slow SD card?)

src/downloadextractthread.cpp
```diff
@@ -438,6 +427,17 @@ void DownloadExtractThread::extractMultiFileRun()
...
+            // Get both hashes: compressed cache file and uncompressed image data
+            QByteArray cacheFileHash = _cachehash.result().toHex();
+            
+            qDebug() << "Cache file created:";
+            qDebug() << "  Image hash (uncompressed):" << computedHash;
+            qDebug() << "  Cache file hash (compressed):" << cacheFileHash;
+            
+            // Emit both hashes for proper cache verification
+            emit cacheFileHashUpdated(cacheFileHash, computedHash);
+            // Keep old signal for backward compatibility
```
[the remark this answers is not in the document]
tdewey-rpi: There are a range of call sites still expecting to use the old "cacheFileUpdated" signal. I've not yet corrected all of them to use the new API.

src/driveformatthread.cpp
```diff
@@ -1,215 +1,259 @@
...
+
+    qDebug() << "Formatting device" << _device << "with clean-room implementation";
+
+    // Get device size
+    std::uint64_t deviceSize = getDeviceSize(_device);
+    qDebug() << "Device size:" << deviceSize << "bytes";
+
+    // Use our clean-room disk formatter
+    rpi_imager::DiskFormatter formatter;
+    auto result = formatter.FormatDrive(_device.toStdString());
```
lurch, line 223: I wonder if it might be worth trying to refactor this multi-OS formatting code so that `formatter.FormatDrive` only appears in one place in `DriveFormatThread::run`? Also, AFAICT (from reading this diff) the "clean-room disk formatter" doesn't seem to be used on macOS?

src/driveformatthread.cpp
```diff
@@ -1,215 +1,259 @@
...
+
+    if (!result) {
+        QString errorMessage;
+        switch (result.error()) {
+            case rpi_imager::FormatError::kFileOpenError:
+                errorMessage = tr("Error opening device for formatting");
```
lurch, line 229: These error-message strings appear in multiple places - perhaps they ought to be in one central place? (just to avoid the possibility of them drifting out of sync)

src/file_operations.h
```diff
@@ -0,0 +1,58 @@
...
+  kSuccess,
+  kOpenError,
+  kWriteError,
+  kSeekError,
+  kSizeError,
+  kCloseError,
+  kLockError
+};
+
+// Abstract interface for platform-specific file operations
```
lurch, line 26: Presumably none of this is duplicating already-existing Qt functionality?

tdewey-rpi: Avoiding duplicating Qt functionality is a non-goal. New code that can avoid relying on Qt should avoid doing so.

src/imagewriter.cpp
```diff
@@ -334,19 +358,55 @@ void ImageWriter::startWrite()
 
     if (!_expectedHash.isEmpty() && _cachedFileHash == _expectedHash)
     {
-        // Check cache file integrity before using it
-        if (_verifyCacheFileIntegrity())
+        // Use background cache manager to check cache file integrity
```
lurch, line 361: This might be a stupid question, but does it ignore all the cache stuff if you select a local already-downloaded `.img` file?

src/imagewriter.h
```diff
@@ -209,7 +222,7 @@ protected slots:
 protected:
     QUrl _src, _repo;
     QString _dst, _cacheFileName, _parentCategory, _osName, _currentLang, _currentLangcode, _currentKeyboard;
-    QByteArray _expectedHash, _cachedFileHash, _cmdline, _config, _firstrun, _cloudinit, _cloudinitNetwork, _initFormat;
+    QByteArray _expectedHash, _cachedFileHash, _cacheFileHash, _cmdline, _config, _firstrun, _cloudinit, _cloudinitNetwork, _initFormat;
```
lurch, line 225: One might argue that having two member-variables with only a one-character difference (`_cachedFileHash` and `_cacheFileHash`) might make them easy to confuse?  :laughing:

src/downloadextractthread.cpp
```diff
@@ -554,10 +570,74 @@ void DownloadExtractThread::_pushQueue(const char *data, size_t len)
...
+    qDebug() << "Verify hash:" << _verifyhash.result().toHex();
+    qDebug() << "Verify done in" << t1.elapsed() / 1000.0 << "seconds";
+
+    if (_verifyhash.result() == _writehash.result() || !_verifyEnabled || _cancelled)
+    {
+        return true;
+    }
+    else
+    {
+        DownloadThread::_onDownloadError(tr("Verifying write failed. Contents of SD card is different from what was written to it."));
```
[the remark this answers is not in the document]
lurch: Fair enough.

touhidurrr:
Can you merge #1088 quick and regenerate translations so that I can commit the changes required by this patch before release so that Bangla translation is dropped at 100%?

touhidurrr:
@tdewey-rpi, since #1088 is merged now can you kindly pull it to this branch and regenerate translations so that i can translate missing lines and pr into this branch.

lurch:
:tada:

PR #1109, opened 2025-07-11
Added Support for Device Detection when in Embedded Mode

bebon901:
rpi-imager will now attempt to read /proc/cpuinfo to determine the model of Raspberry Pi Device you are running on. If the device is running the Imager in embedded mode, then it shall restrict the hardware options to the device that Imager is running on.


src/linux/device_info.cpp
```diff
@@ -0,0 +1,127 @@
...
+            QString line = s.readLine();
+            if (line.startsWith("Revision")) {
+                _revision = line.section(':', 1).trimmed();
+                bool ok;
+                uint32_t value_int = _revision.toUInt(&ok, 16);
+
+                // Device type is 8 bits, stored in bits 4-11
+                const uint32_t MASK = 0xFF << 4;
+                uint32_t isolated_bits = value_int & MASK;
+                uint32_t extracted_value = isolated_bits >> 4;
```
lurch, line 34: I guess this is assuming that the device is using a new-style revision code. Should we also support old-style revision codes, or do they not matter as those devices probably won't be able to run the Embedded Imager anyway?
https://www.raspberrypi.com/documentation/computers/raspberry-pi.html#raspberry-pi-revision-codes
At the very least, you should probably check that the "New flag" (bit 23) is set, before trying to read bits 4-11 ?

src/linux/device_info.cpp
```diff
@@ -0,0 +1,127 @@
...
+                    {0x19, "Raspberry Pi 5"},
+                    {0x1a, "Raspberry Pi 5"}
+                };
+                try
+                {
+                    _hardware_name = QString::fromStdString(pi_revision_map[extracted_value]);
+                    _is_raspberry_pi = true;
+                    return true;
+                }
+                catch(const std::exception& e)
```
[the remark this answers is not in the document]
lurch: It _looks_ like the only thing being caught is `extracted_value` not appearing in the `pi_revision_map` ?

src/linux/device_info.cpp
```diff
@@ -0,0 +1,127 @@
...
+                    {0x0a, "Raspberry Pi 3"},
+                    {0x0c, "Raspberry Pi Zero"},
+                    {0x0d, "Raspberry Pi 3"},
+                    {0x0e, "Raspberry Pi 3"},
+                    {0x10, "Raspberry Pi 3"},
+                    {0x11, "Raspberry Pi 4"},
+                    {0x12, "Raspberry Pi Zero 2 W"},
+                    {0x13, "Raspberry Pi 4"},
+                    {0x14, "Raspberry Pi 4"},
+                    {0x15, "Raspberry Pi 4"},
```
lurch, line 61: Just for reference and readability purposes, it might be nice to have comments here clarifying the _actual_ Pi model? E.g.
```suggestion
                    {0x13, "Raspberry Pi 4"}, // 400
                    {0x14, "Raspberry Pi 4"}, // CM4
                    {0x15, "Raspberry Pi 4"}, // CM4S
```

src/linux/device_info.cpp
```diff
@@ -0,0 +1,127 @@
...
+    return false;
+}
+
+bool device_info::setHardwareTags(QJsonArray deviceArray)
+{
+    // At this point, the OS List has been retrieved, a signal has been sent to run this section
+    // Also a good time to determine whether or not this is a Raspberry Pi Device
+    determineHardware(); 
+        // Determine the correct filtering tags for use within the OSListFilter
+        for (const QJsonValue &deviceValue: deviceArray) {
```
lurch, line 93: Mismatched indentation?

bebon901:
Worth noting that I haven't actually been able to test this on an embedded pi since the changes were made!

src/imagewriter.h
```diff
@@ -61,6 +63,14 @@ class ImageWriter : public QObject
     /* Returns true if src and dst are set */
     Q_INVOKABLE bool readyToWrite();
 
+    /* Returns true if running on a Raspberry Pi */
+    Q_INVOKABLE bool isRaspberryPiDevice();
+
+    /* Returns true if running on a Raspberry Pi */
```
lurch, line 69: Copy'n'paste comment? (Seems unlikely that a function returning a `QString` will "return true" :wink: )

lurch:
Looks like your rebase has gone awry?

PR #1119, opened 2025-07-17
Update error in README.md

itsMattShull:
Under macOS there was a Windows reference


lurch:
The backslashes on line 91 should probably be forward-slashes too?

lurch:
> The backslashes on line 91 should probably be forward-slashes too?

Created #1122 for that.

PR #1132, opened 2025-07-21
Embedded Imager - Add Device Detection

bebon901:


src/embedded/device_info.cpp
```diff
@@ -0,0 +1,137 @@
...
+                    _revision = line.section(':', 1).trimmed();
+                    bool ok;
+                    uint32_t value_int = _revision.toUInt(&ok, 16);
+                    if (!ok) { qDebug() << "Revision Code Not OK!"; }
+
+                    // New Style Revision Code
+                    // Device type is 8 bits, stored in bits 4-11
+                    const uint32_t MASK = 0xFF << 4;
+                    uint32_t isolated_bits = value_int & MASK;
+                    uint32_t extracted_value = isolated_bits >> 4;
```
lurch, line 34: Doesn't really matter, but I guess this could be shortened to just:
```c
uint32_t extracted_value = (value_int >> 4) & 0xFF;
```
?

src/embedded/device_info.cpp
```diff
@@ -0,0 +1,137 @@
...
+
+                    // convert final value back to HEX
+                    std::stringstream ss;
+                    ss << std::hex << extracted_value;
+                    QString extracted_hex_string = QString::fromStdString(ss.str());
+                    _device_code = extracted_hex_string;
```
lurch, line 40: I think I'd be tempted to store the `_device_code` as the actual integer, and only convert to hex-string when needed, rather than storing `_device_code` as a `QString`? But of course that's entirely up to you.

src/embedded/device_info.cpp
```diff
@@ -0,0 +1,137 @@
...
+
+                    // Make a Human Readable Equivalent of this device_code
```
lurch, line 42: Slightly random CapiTaliSation? :wink:

src/CMakeLists.txt
```diff
@@ -4,6 +4,7 @@
 cmake_minimum_required(VERSION 3.22)
 OPTION (ENABLE_CHECK_VERSION "Check for version updates" ON)
 OPTION (ENABLE_TELEMETRY "Enable sending telemetry" ON)
+OPTION (BUILD_EMBEDDED "Build for Embedded Imager" OFF)
```
tdewey-rpi, line 7: Do we also need the change to create-embedded.sh to use this?

src/embedded/device_info.cpp
```diff
@@ -0,0 +1,137 @@
...
+        QFile f("/proc/cpuinfo");
+        if (f.open(QFile::ReadOnly | QFile::Text)) {
+            QTextStream s(&f);
+            while (!s.atEnd()) {
+                QString line = s.readLine();
+                if (line.startsWith("Revision")) {
+                    _revision = line.section(':', 1).trimmed();
+                    bool ok;
+                    uint32_t value_int = _revision.toUInt(&ok, 16);
+                    if (!ok) { qDebug() << "Revision Code Not OK!"; }
```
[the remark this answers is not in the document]
tdewey-rpi: Agreed. This feels like an early-return and set an error state location.

src/main.qml
```diff
@@ -849,7 +849,28 @@ ApplicationWindow {
...
+                    // Limit the hardware selection button to only the specified device
+                    text0.text = "Hardware Type Detected";
+                    hwbutton.enabled = false
+                    hwbutton.text = hardwareTag;
+                }
+            }
+        }
+        else {
+            console.log("Seems we're not a raspberry pi...");
+        }
```
tdewey-rpi, line 873: Could this be radically simplified - such that we just call getHardwareName, and if the result is a non-zero-length string, set that as the hardware tag?

src/CMakeLists.txt
bebon901: Have put changes to create-embedded.sh in separate PR that addresses how rpi-imager-embedded is built

bebon901:
Thank you for all the comments @lurch.
Have corrected all of the above.

The code is now also fully operational - it now restricts hardware options when operating in embedded mode.

lurch:
Looks much better now @bebon901 , thanks for taking all my feedback on board. However thinking about "separation of concerns", I wonder if all the hardware-tags stuff would be better being just inside `imagewriter`, rather than also being inside `device_info` ?

src/main.qml
```diff
@@ -880,6 +889,8 @@ ApplicationWindow {
         else if (imageWriter.isEmbeddedMode()) {
             //msgpopup.text = qsTr("<b>%1</b> has been written to <b>%2</b>").arg(osbutton.text).arg(dstbutton.text)
             /* Just reboot to the installed OS */
+            msgpopup.text = qsTr("<b>%1</b> has been written to <b>%2</b><br><br>You can now unplug your Raspberry Pi")
```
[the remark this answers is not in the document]
lurch: Ahhh, I've not looked at that other issue yet :laughing:

src/config.h
```diff
@@ -11,7 +11,7 @@
 #define OSLIST_URL                        "https://downloads.raspberrypi.org/os_list_imagingutility_v4.json"
 
 /* Time synchronization URL (only used on eglfs QPA platform, URL must be HTTP) */
```
tdewey-rpi, line 13: Maybe update this comment, too.

tdewey-rpi (approved):

PR #1134, opened 2025-07-21
Translation into Russian for another field

nafanz:
I hope there won't be any new fields in the near future ;-)


tdewey-rpi:
Unfortunately, I'm adding new fields at a fair pace - mostly to expose users to the workings of the application. Expect more time between the next release, however.

tdewey-rpi (approved):

PR #1135, opened 2025-07-21
windows: Refactor drive formatting

tdewey-rpi:
- Always use diskpart clean ahead of time
- Fail if diskpart fails
- Plumb through formatting messages to the UI
- Add a 60s timeout if it stalls


src/downloadthread.cpp
```diff
@@ -155,23 +160,92 @@ bool DownloadThread::_openAndPrepareDevice()
...
+                    qDebug() << "diskpart succeeded on attempt" << attempt;
+                    break;
+                }
+                else
+                {
+                    QString errorOutput = QString(proc.readAllStandardError());
+                    qDebug() << "diskpart failed on attempt" << attempt << "with error:" << errorOutput;
+                    
+                    // Wait before retry
+                    QThread::msleep(1000 * attempt); // Progressive delay
```
lurch, line 236: As a micro-optimisation I guess you could skip the 3-second sleep when `attempt == 3`, as there aren't going to be any more retries?

src/driveformatthread.cpp
```diff
@@ -63,8 +67,67 @@ void DriveFormatThread::run()
...
+        {
+            emit error(tr("Failed to start disk cleanup utility. Please ensure you have administrator privileges."));
+            return;
+        }
+        
+        QString script = QString("select disk %1\nclean\nrescan\n").arg(diskNumber);
+        diskpartProcess.write(script.toLatin1());
+        diskpartProcess.closeWriteChannel();
+        
+        if (!diskpartProcess.waitForFinished(60000))
```
lurch, line 109: Different timeout value here (60 seconds) to in `downloadthread.cpp` (30 seconds)?
This is probably a stupid question, but why does `downloadthread.cpp` retry diskpart up to 3 times, but `driveformathread.cpp` only does it once?
Even stupider question: I wonder if it might be worth doing a small refactoring so that the diskpart-related code isn't "duplicated" in different places?

tdewey-rpi:
Tested on Windows 11. Erasing working as expected.

PR #1140, opened 2025-07-22
Changes to fix how embedded_imager is created

bebon901:
The new build-qt-embedded now will omit a large amount of features irrelevant to building Qt for the embedded imager.
Added service file to start embedded imager upon boot on embedded systems.
Changes to create-embedded to render using linuxfb, instead of EGLFS


tdewey-rpi:
Definitely headed in the right direction. One comment-grade change to make.

PR #1141, opened 2025-07-23
Updating the list of Pi models when creating a bug

nafanz:
I assume that 500, and Compute Modules 5 were unintentionally forgotten.

The Raspberry Pi 2B is of course an old model, but some users (like me) still use it.

I don't remember if there were other variants of the Raspberry Pi 2, please correct me if I missed something.

Off topic, but I wonder if the future Raspberry Pi OS (Debian 13, Trixie) will still support the 32-bit architecture and therefore the Raspberry Pi 2?

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug-report.yml
+++ b/.github/ISSUE_TEMPLATE/bug-report.yml
@@ -72,10 +72,11 @@ body:
       label: Which Raspberry Pi Device are you using?
       multiple: false
       options:
-        - Raspberry Pi 5
+        - Raspberry Pi 5, 500, and Compute Modules 5
         - Raspberry Pi 4B, 400, and Compute Modules 4, 4S
         - Raspberry Pi Zero 2 W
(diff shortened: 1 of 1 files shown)


nafanz:
I remembered that you can check with `https://downloads.raspberrypi.com/os_list_imagingutility_v4.json`

It turns out that everything is correct for Raspberry Pi 2 there was only Model B.

```
{
    "name": "Raspberry Pi 2",
    "tags":
    [
        "pi2-32bit"
    ],
    "default": false,
    "icon": "https://downloads.raspberrypi.com/imager/icons/RPi_2.png",
    "description": "Model B",
    "matching_type": "inclusive"
}
```

lurch:
> Off topic, but I wonder if the future Raspberry Pi OS (Debian 13, Trixie) will still support the 32-bit architecture and therefore the Raspberry Pi 2?

Yes, Raspberry Pi OS Trixie will continue our tradition of supporting **every** model of Raspberry Pi :grinning:

PR #1149, opened 2025-07-29
Add openUrl method to ImageWriter for external URL handling

tdewey-rpi:
Implemented a new Q_INVOKABLE method in ImageWriter to open URLs using platform-specific commands (xdg-open for Linux, open for macOS, and start for Windows). Added fallback to QDesktopServices if the platform command fails. Updated main.qml to utilize the new method for opening URLs externally.


lurch:
Included as part of #1135

PR #1181, opened 2025-08-27
Fix Windows build

paulober:
Fixes the libz linking / compilation and the timezone.txt generation.


src/qml.qrc
```diff
@@ -4,7 +4,6 @@
     <qresource prefix="/">
         <file>qtquickcontrols2.conf</file>
         <file alias="countries.txt">countries.txt</file>
-        <file>timezones.txt</file>
```
lurch, line 7: ?? Is there some other process that adds `timezones.txt` back to `qml.qrc`? I think it might be needed by the embedded imager? ping @bebon901

Issue #1190, opened 2025-08-31
[BUG]: rpi-imager offers to use invalid WiFi Regulatory Domain Codes


jvonau:
### What happened?

User selected 'CD' for the WiFi Regulatory Domain Code https://github.com/iiab/iiab/issues/3890#issuecomment-2594691466 but that country code is not part of  https://git.kernel.org/pub/scm/linux/kernel/git/wens/wireless-regdb.git/plain/db.txt causing a bunch of issues.
https://github.com/iiab/iiab/issues/3912#issue-2808191547

### Version

older version issue was found

### What host operating system were you using?

Debian and derivatives (eg Ubuntu)

### Host OS Version

U24.04

### Selected OS

RasPi OS Bookworm 64bit desktop

### Which Raspberry Pi Device are you using?

Raspberry Pi 4

### What kind of storage device are you using?

microSD Card in an internal reader

### OS Customisation

- [x] Yes, I was using OS Customisation when the bug occurred.

jvonau:
The above info about Version is not accurate as this issue was found many months ago but an issue was never filed by @holta while https://github.com/raspberrypi/rpi-imager/blob/qml/src/countries.txt hasn't been updated in 3 years
https://github.com/iiab/iiab/issues/3912#issuecomment-2611334688

tdewey-rpi:
Fixed in 2.0.

PR #1193, opened 2025-09-01
Use wireless-regdb.git's db.txt as the as the authoritative source for WiFi Regulatory Domain Codes

jvonau:
Use https://git.kernel.org/pub/scm/linux/kernel/git/wens/wireless-regdb.git/tree/db.txt as the as the authoritative source
for the WiFi Regulatory Domain Codes
Prepared with:
`wget https://git.kernel.org/pub/scm/linux/kernel/git/wens/wireless-regdb.git/tree/db.txt`
`grep country db.txt | sed s"|country||" | sed s"|:||" | awk '{print $1}' | sed /00/d > db.clean`
`diff -u src/countries.txt db.clean > countries.patch`
`patch -u -i countries.patch  src/countries.txt`


jvonau:
https://github.com/raspberrypi/rpi-imager/pull/1190

jedahan:
Would be cool to add that preparatory script somewhere in a release or contrib folder for next time it should be updated?

lurch:
> Would be cool to add that preparatory script somewhere in a release or contrib folder for next time it should be updated?

Yes, it's very reminiscent of https://github.com/raspberrypi/rpi-imager/blob/qml/src/cmake/GenerateTimezones.cmake :slightly_smiling_face:

jvonau:
Better yet would be to use a list of countries that are actually supported by the firmware (I don't have access to this information) as using HT that is found in wireless-regdb but results in `ieee80211 phy0: brcmf_cfg80211_reg_notifier: Firmware rejected ` in dmesg https://github.com/iiab/iiab/issues/3868#issuecomment-2594315300

lurch:
> Better yet would be to use a list of countries that are actually supported by the firmware

That's _entirely_ dependent on the version of the WiFi firmware being used in the about-to-be-flashed OS image, and there's obviously no way for Raspberry Pi Imager to know that.

jvonau:
Any idea on where to find this information or is this a NDA situation?

lurch:
The WiFi firmware for Raspberry Pi OS comes from this repo https://github.com/RPi-Distro/firmware-nonfree but of course other OSes may use different firmware? :shrug:  I'm afraid that it's not something that I know much about.

jvonau:
> Better yet would be to use a list of countries that are actually supported by the firmware (I don't have access to this information) as using HT that is found in wireless-regdb but results in `ieee80211 phy0: brcmf_cfg80211_reg_notifier: Firmware rejected ` in dmesg [iiab/iiab#3868 (comment)](https://github.com/iiab/iiab/issues/3868#issuecomment-2594315300)

@pelwell @waveform80 any ideas?

pelwell:
I've not been through every edit but the methodology seems sound and the result looks correct.

tdewey-rpi:
Thanks for the submission, @jvonau 

Unfortunately, it's already implemented for -next: https://github.com/raspberrypi/rpi-imager/blob/3bd927f42bc31d2889be7c0ed6ca4795e7a84d2d/src/CMakeLists.txt#L776

jvonau:
> Thanks for the submission, @jvonau
> 
> Unfortunately, it's already implemented for -next:
> 
> https://github.com/raspberrypi/rpi-imager/blob/3bd927f42bc31d2889be7c0ed6ca4795e7a84d2d/src/CMakeLists.txt#L776

Thanks for that.

PR #1194, opened 2025-09-01
OptionsPopup: duplication of ssh keys, making the interface unusable …

alinsalaru:
…after a period of time

This means that every time the dialog opens, it appends the saved keys to whatever keys are already in the model, bloating the whole app.

### DIFF
--- a/src/OptionsPopup.qml
+++ b/src/OptionsPopup.qml
@@ -534,6 +534,8 @@ Window {
         }
         if ('sshAuthorizedKeys' in settings) {
             var possiblePublicKeys = settings.sshAuthorizedKeys.split('\n')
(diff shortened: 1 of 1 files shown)


alinsalaru:
@tdewey-rpi really appreciate the work you've done for this project! 
Can you please look at into this? it's a small change but offers a good performance improvement.

All the best in the world!

lurch:
The `qml` branch corresponds to the 1.9.x releases of RPi Imager, but I believe that active-development of Imager has now switched to the `main` branch (for the upcoming 2.0.x releases).

alinsalaru:
Indeed, my mistake, will follow on main branch. 
And close this, thank you @lurch

PR #1204, opened 2025-09-09
Improve debian package multi-arch build

mhalano:
Improve the Debian package build to use multiple architectures correctly.

### DIFF
--- a/debian/control
+++ b/debian/control
@@ -7,24 +7,13 @@ Standards-Version: 4.1.2
 Homepage: https://www.raspberrypi.com/software
 
 Package: rpi-imager
-Build-Profiles: <!amd64 !embedded>
-Architecture: arm64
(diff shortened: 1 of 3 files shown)


mhalano:
@tdewey-rpi I tested on my AMD64. The change to compile is using `-a amd64` instead of `--build-profiles=amd64`.
You can see it's an elegant solution.
I will test on my Raspberry Pi 4, but since I need to compile all Qt and then generate the AppImage, that could take some time.
If you can test faster, please do so.

tdewey-rpi:
@mhalano That looks pretty good, thanks for the PR!

I'd ask that you rebase it against the 'main' branch, so we can release 2.0.0 in this form.

mhalano:
@tdewey-rpi Rebased against main.

PR #1207, opened 2025-09-10
Add local repository selector to settings

paulober:
Currently only supports local repository files.

### DIFF
--- a/src/CommonStrings.qml
+++ b/src/CommonStrings.qml
@@ -14,24 +14,28 @@ Item {
     readonly property string systemDriveText: qsTr("System drives typically contain files essential to the operation of your computer, and may include your personal files (photos, videos, documents).")
 
     // --- New reusable filters ---
+    readonly property string allFilesLabel: qsTr("All files (*)")
+    readonly property var    allFilesList: [ allFilesLabel ]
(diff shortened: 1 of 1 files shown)


lurch:
Is this basically a graphical equivalent of the `--repo` CLI flag? Do we expect this feature to be used often enough to warrant adding it to the GUI?

paulober:
> Is this basically a graphical equivalent of the `--repo` CLI flag?

Yes, it is, and it can be prefilled with the `--repo` argument.

src/wizard/RemoteAccessStep.qml
```diff
@@ -181,7 +181,7 @@ WizardStepBase {
     ImFileDialog {
         id: sshKeyFileDialog
         dialogTitle: qsTr("Select SSH Public Key")
-        nameFilters: ["*.pub", "*"]
+        nameFilters: ["Public Key files (*.pub)", "All files (*)"]
```
tdewey-rpi, line 184: Reuse commonstrings.allFileFilter here?

tdewey-rpi:
Good start - but factoring out the filters should happen before it goes in.

tdewey-rpi:
Approved!

PR #1210, opened 2025-09-12
Fix locale loading on Windows

paulober:


tdewey-rpi:
While I would have preferred this without the formatting changes, I'll accept as-is.

paulober:
> While I would have preferred this without the formatting changes, I'll accept as-is.

Didn't want to be petty, that was just the auto-format of the editor.

PR #1175, opened 2025-08-21
Allow to build without vendoring

doronbehar:
### Commits:

- **CMake: remove now not used nghttp2 dep**
- **Try to fetch dependencies from system first**

### Purpose:

Make it easier for Arch Linux, Nix, and possibly others to distribute your package with the system installed dependencies.

PR #1211, opened 2025-09-13
CMake: remove now not used nghttp2 dep

doronbehar:
Was part of https://github.com/raspberrypi/rpi-imager/pull/1175 .


lurch:
I believe active-development is now happening on the `main` branch, so could you rebase against that please?

doronbehar:
> I believe active-development is now happening on the `main` branch, so could you rebase against that please?

Done!

tdewey-rpi:
Huh?

curl.cmake: set(USE_NGHTTP2 ON CACHE BOOL "" FORCE)

Why do you believe NGHTTP2 isn't being used, @doronbehar?

doronbehar:
Hmm OK I see. So since you build curl with nghttp2 support, you also need to build nghttp2. Because you don't use nghttp2 directly, I didn't notice this dependency is needed since on NixOS we insist using our dependencies. Our curl is built with nghttp2 so we haven't had any issues between rpi-imager and nghttp2.

PR #1218, opened 2025-09-23
feat: translate to Polish

Omikorin:
I did everything, but I needed to make some workarounds.

The window is too thin for some texts in Polish as Polish language is more descriptive than English, even though I tried to minimize unneeded words. I think we could make the window a bit wider and use more `<br>` elements for these small captions under inputs.

Examples below:

---

"Pomiń dostosowywanie" as "skip customization":
The button is the only culprit for extending input's width.

hostname

"Pomiń" as "skip":

hostname2

---

The input's text is too long.

ssh

I went with solution below. I think it was too descriptive anyway, because it should be enough for user to tell him "insert key" and the next thing is a button with "browse" instead of descriptive (repetitive) input text. (knowledge from my frontend journey).

ssh2

---

Moreover, the "Customization: choose hostname" and similar are too long in Polish language, so I removed the "customization:" part. These texts are marked as `unfinished` to check them someday.

I don't know how you handle untranslatable strings like MB, GB, KB, B, raspberrypi (the hostname thing) etc. so I kept them unchanged. Maybe let's mark their type as `untranslatable`? I've seen this solution in other projects before.

I found that there is no option to translate the "no filtering" option in devices' selector.

Cheers

### DIFF
--- a/src/i18n/rpi-imager_pl.ts
+++ b/src/i18n/rpi-imager_pl.ts
@@ -5,230 +5,230 @@
     <name>AppOptionsDialog</name>
     <message>
         <source>App Options</source>
-        <translation type="unfinished">Ustawienia aplikacji</translation>
+        <translation>Ustawienia aplikacji</translation>
     </message>
     <message>
         <source>Play sound when finished</source>
-        <translation type="unfinished">Odegraj dźwięk po zakończeniu</translation>
+        <translation>Odegraj dźwięk po zakończeniu</translation>
     </message>
     <message>
         <source>Eject media when finished</source>
-        <translation type="unfinished">Wysuń nośnik po zakończeniu</translation>
+        <translation>Wysuń nośnik po zakończeniu</translation>
     </message>
     <message>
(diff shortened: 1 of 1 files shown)


tdewey-rpi:
Thanks for the submission, @Omikorin

tdewey-rpi:
> I don't know how you handle untranslatable strings like MB, GB, KB, B, raspberrypi (the hostname thing) etc. so I kept them unchanged. Maybe let's mark their type as untranslatable? I've seen this solution in other projects before.

By convention, we're using GB to represent GiBiBytes (1024, rather than 1000). In French, they prefer Gigaoctet to Gigabyte, so they use GO instead of GB. For many languages, I don't expect these units to be displayed differently - but Imager 2.0 becomes the first version where France can have more of the UI rendered correctly.

> I found that there is no option to translate the "no filtering" option in devices' selector.

That's because 'No filtering' is not specified in the application - but in the JSON instead. I'd expect this translation to be handled by a _pl.json repository file - but that's not been prepared as yet.

Thanks again for the submission - verified with machine translation.

Issue #209, opened 2021-04-13
Provide a simple way to delete cached image


ralphrmartin:
If you only want to use this to set up one Raspberry Pi, it makes no sense to keep the cached download, which can waste a significant amount of disk space. It is not obvious to users that the file is being cached - I assumed it would be deleted on exit (or perhaps be in /tmp). Please provide some simple user interface for deleting it, e.g. a checkbox "Remove download after installing".

maxnet:
But if you only ever want to use it to setup that single Pi, does it make sense to keep Raspberry Pi Imager installed?
The uninstaller does remove it...

>I assumed it would be deleted on exit (or perhaps be in /tmp)

Technically the program does not need any temporary files to function. Can stream files from the Internet straight to SD card, uncompressing on the fly.
Caching was added later.

Problem is the UI needs to beginner friendly, leaving out configurable options.

ralphrmartin:
The standard, Apple way, to uninstall stuff on a Mac is to drag the application to the trash, so that won't delete the cache. It does not have uninstallrs like Windows and Linux.

I hardly think it's going to confuse even beginners to say "Keep a copy of the downloaded Raspbian OS for further installations in future".

I'd say its even less friendly to silently waste a few GB, especially for users with small disks.

maxnet:
>The standard, Apple way, to uninstall stuff on a Mac is to drag the application to the trash, so that won't delete the cache.

Yeah, that do is a hard to solve problem.

Apple only cleans up properly for apps that come from Mac Store.
But our application is not welcome there (they do not allow any applications that needs to do anything privileged there...)

alfieroddan:
I can't actually find the cached image on my Mac. 

Is there a standard location for the Rasperry Pi Imager to store images? 

I agree with @ralphrmartin that even a simple delete cache on the UI would be helpful. I am said "beginner" on booting different images on to my pi. 

Love the app so far tho.

lurch:
I think RPi Imager only caches the last-flashed image, so it doesn't matter if you flash multiple different images onto your SD card. (so I guess a sneaky way to reduce the amount of disk-space used by the cache would be to flash one of the smaller OSes? :wink: )

maxnet:
>I can't actually find the cached image on my Mac.

Menu "Go" -> "Go to Folder..." -> ~/Library/Caches/Raspberry Pi

raisaab198:
[...]

PR #1220, opened 2025-09-24
FIX Set cache in temp folder

ThibaultDECO:
In order to make it easier for users to clean the cache, this PR sets cache in the temp folder.

This fixes issue #209

### DIFF
--- a/src/cachemanager.cpp
+++ b/src/cachemanager.cpp
@@ -434,7 +434,9 @@ void CacheManager::saveCacheSettings()
 
 QString CacheManager::getDefaultCacheFilePath() const
 {
-    return QStandardPaths::writableLocation(QStandardPaths::CacheLocation) + 
+    return QStandardPaths::writableLocation(QStandardPaths::TempLocation) + 
+           QDir::separator() + QCoreApplication::applicationName() + 
+           QDir::separator() + "cache" + 
            QDir::separator() + "lastdownload.cache";
 }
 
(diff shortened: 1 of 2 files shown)


lurch:
Would this cause a problem if somebody has set their temp directory to be in RAM, but then they try to flash an image that's bigger than their available RAM size?

ThibaultDECO:
@lurch Hi!
I'm not sure I understand what your concern is since:

- This is already the case for the cache directory: if someone explicitly points it to RAM and the file is bigger than available RAM, then yes, it would fail, just like if it were pointed to a disk with insufficient space. More generally, if cache or temp is set to a location with less space than the image size, it won't work, regardless of whether that location is RAM or a disk.
- The Imager doesn't need to load the whole image in RAM: it streams data in chunks, so only a small buffer is ever held in RAM.
- Since this is the Imager, I assume the vast majority of people use a "standard" computer to flash the image, therefore not running on RAM.
- For edge cases, the vast majority of devices running on RAM nowadays have RAM well above the 1GB required for an image.

tdewey-rpi:
PR rejected.

The concerns raised by @lurch are incompatible with this approach - memory usage is absolutely a concern, and this patch would also affect the embedded Imager running on Raspberry Pi devices.

Unfortunately, this PR highlights why Imager is an extraordinarily complex beast to develop - and why releases take longer than some expect.

PR #1224, opened 2025-09-29
New repo source editing dialog

paulober:


tdewey-rpi:
This wants a rebase now, but the concept is approved.

paulober:
I do the apply settings in the repo dialog because it has the context but then you also have the save button in the parent dialog and that's maybe confusing?
So maybe make the edit button be named something with the meaning switch to repo dialog and perform an apply settings for the main app options before and then after the repo dialog finished it just closes and doesn't go back to app options...

PR #1225, opened 2025-09-29
Prework for removing rpios cloud-init customization

paulober:


tdewey-rpi:
Not convinced we want this - but we may want to turn on these later via a schema modification?

PR #1227, opened 2025-10-01
Integrate DeviceWrapper with new fileops

paulober:


tdewey-rpi:
Alternative patch chosen - thanks for the PR!

PR #1239, opened 2025-10-07
Complete Simplified Chinese translation

ykla:
Add Simplified Chinese translation for SSH key error.


ykla:
It seems that I chose the wrong branch.

PR #1240, opened 2025-10-07
Complete translation into Simplified Chinese

ykla:
Why are there so many duplicate sentences? Can it be converted into a `.pot` file?

### DIFF
--- a/src/i18n/rpi-imager_zh.ts
+++ b/src/i18n/rpi-imager_zh.ts
@@ -5,230 +5,230 @@
     <name>AppOptionsDialog</name>
     <message>
         <source>App Options</source>
-        <translation type="unfinished"></translation>
+        <translation>应用选项</translation>
     </message>
     <message>
         <source>Play sound when finished</source>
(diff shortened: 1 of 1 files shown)


tdewey-rpi:
> Why are there so many duplicate sentences?

I've now made changes to reduce the number of duplicates - I had hoped the Qt tooling would've picked up on this, but it hasn't.

Unfortunately I don't think we can migrate to .pot files without a major change through the rest of the program.

lurch:
You might want to rebase this against `dev/tdewey/2.0.2-refinement` to ensure that you've got all the strings added by #1314 ?

ykla:
> You might want to rebase this against `dev/tdewey/2.0.2-refinement` to ensure that you've got all the strings added by #1314 ?

Since the Raspberry Pi can no longer be maintained on FreeBSD, I have stopped using it. I now see that others are attempting to maintain these translations, and I am willing to hand it over to them. However, if in the end they do not complete it or do not continue, please @ me, and I will still take care of it. For now, I think it's time to close this PR first.

Issue #1188, opened 2025-08-30
[BUG]: System can go to sleep, and when it does, image download/write is interrupted


logiclrd:
### What happened?

I fired up the Raspberry Pi Imager to write an image for use on a device, and the Internet connection was really slow. I left the system to do its thing and when I returned some time later, I found that it had decided it was idle and had gone to sleep. When it came out of sleep mode, the operation sat for a long time waiting for the next network packet that was never going to come and then failed with an error.

* The tool should make use of platform-specific facilities to disable automatic suspend during the execution of a DownloadThread.
* If the system _does_ go to sleep, a more robust retry/error-handling strategy is needed. For instance, instead of simpy aborting, the user could be prompted about whether to try to resume the process if automated retry has failed too many times. The time it takes for the user to react may give Wi-Fi enough time to reconnect.

I found this cross-platform C++ Gist showing how to inhibit suspend on Windows, OS X and Linux (KDE and Gnome):

https://gist.github.com/Vineg/eca223fbf478a3c806444a13e538a9fc

### Version

1.9.6 (Default)

### What host operating system were you using?

Windows

### Host OS Version

Windows 10

### Selected OS

Raspberry Pi OS 64-bit

### Which Raspberry Pi Device are you using?

Raspberry Pi Zero 2 W

### What kind of storage device are you using?

microSD Card in an internal reader

### OS Customisation

- [x] Yes, I was using OS Customisation when the bug occurred.

### Relevant log output

```shell

```

yanxke:
Should just resume if the network is interrupted.  Switching Wifi networks will also cause this.  Modern software should be reliable against these simple failures.

logiclrd:
It didn't resume, but also if it's a time-consuming process, I want to be able to walk away and come back some time later and not find that the computer went to sleep after 5 minutes and nothing further has happened.

logiclrd:
I suspect it does have a retry mechanism, but when coming out of sleep mode, code starts running well before the network has reinitialized, so it would quickly exhaust retries with immediate failures due to the network not yet being established. That is what I'm referring to when I say I think it needs a _more robust_ strategy.

tdewey-rpi:
Merged for 2.0. Closing.

PR #1251, opened 2025-10-11
Inhibit suspend during the processing of an imaging operation

logiclrd:
> This pull request is being created as a draft because I have no way to test the code. :-)
> 
> - I don't have a Linux or Mac build environment. (I do have a Linux system but it doesn't have a working Qt installation.)
> - I tried to configure a Windows build environment but messed up somewhere along the way. It's trying to use CL to build the C code, but it's using GNU extensions so that's not working.
> 
> I have executed the Windows and Linux specific inhibitor code in isolated test projects and it builds and seems to work. But, I don't have any way to tell if the other code even compiles. I'm hoping there will be build pipelines that'll take care of at least part of that.
> 
> Someone with a Mac will need to test the Mac code.

This PR is a direct response to a poor user experience I had with the tool. Per issue #1188, I set up an operation to download & write an image to a Micro SD card, and then I left the system doing its thing. When I returned, I found out that the system had deemed itself idle and gone to sleep after 15 minutes, long before the download completed.

This PR aims to address this by leveraging OS-specific functionality to inhibit autosuspend.

- On Windows, it sets the thread execution state to include the flag `ES_SYSTEM_REQUIRED`.
- On OS X, it creates an IOPM assertion of type `PreventUserIdleSystemSleep`.
- On Linux, it tries three strategies:
    - It attempts to connect to the DBus endpoint `org.gnome.SessionManager`. If this succeeds, it makes a call to the `Inhibit` method.
    - It attempts to launch system utility `systemd-inhibit`.
    - It attempts to launch system utility `kde-inhibit`.

These `-inhibit` utilities in the Linux-specific side work by taking a command and executing it in a scope that inhibits whatever action is to be inhibited. The inhibition remains in effect until the process exits. In order to invert this control mechanism, the code here creates a FIFO and then has the command in the `-inhibit` scope read from it. It can then terminate the read at any time by closing its end of the pipe.

Any errors are silently ignored; a best effort was made but if it can't explicitly inhibit suspending, we still want to perform the download & write operation.

### DIFF
--- a/src/windows/Platform.cmake
+++ b/src/windows/Platform.cmake
@@ -1,7 +1,5 @@
 # Windows platform-specific sources and link settings
 
-set(CMAKE_DLLTOOL "C:\\msys64\\mingw64\\bin\\dlltool.exe")
-
 # Find dlltool for creating import library
 if(NOT CMAKE_DLLTOOL)
     find_program(CMAKE_DLLTOOL NAMES dlltool x86_64-w64-mingw32-dlltool)


logiclrd:
Well shucks 🙂 There's no C.I., it seems. Hmm.

tdewey-rpi:
Conceptually approved, but I can't merge it until I've put it through a round of testing.

This feature is sufficiently impactful that it should block 2.0 release.

I expect to come back to this PR on Thursday afternoon at the earliest.

logiclrd:
> a round of testing

Indeed :-) The main reason it was over a month between the issue and the PR is that I hate putting forth code I haven't tested myself. But, ultimately, I struggled with getting a development environment working and still wanted to contribute this feature, so here we are, a draft PR with no end-to-end testing yet. My apologies that that's the best I could muster.

tdewey-rpi:
Broadly happy after initial review.

`catch (...)` patterns make me exceptionally nervous, but in this case the code you're guarding is relatively constrained, and not something where differentiated handling would offer an advantage, so I'm prepared to let it pass by.

tdewey-rpi:
Tested, fixed up, and merged. If you want to see the fixups: https://github.com/raspberrypi/rpi-imager/commit/9d4a8045471c5212a423f8e5975eded26d1dc081

Thanks for a solid improvement!

logiclrd:
Awesome, thanks so much :slightly_smiling_face:

PR #1253, opened 2025-10-14
Fix rpi-connect cb by adding Qt IPC and single-instance on Windows

paulober:
This makes the rpi-connect login callback work on Windows again. It previously failed on Windows because of the UAC and spawning of a new instance.

This now adds Qt IPC via `QLocalServer` and `QLocalSocket` and makes imager single-instance on Windows.


tdewey-rpi:
Happy to merge as-is, but this will become a little bit of tech debt. In the future, we'll want to create a platform hook for this behaviour, to avoid ifdef blocks inside of the main code where we can.

PR #1276, opened 2025-10-21
feat(RpiConnect/Windows): Receive rpi-imager callback without UAC prompt

paulober:


tdewey-rpi:
A good concept in theory - but one we can't accept, because of the ephemeral console window (which at least visually looks like something nefarious is happening).

We'll have to instead implement a helper. As the language of choice for this repository is C++, I'd suggest another C++ binary that can act as the URL handler.

PR #1316, opened 2025-11-26
 Fix: Add timeout protection for fsync/fdatasync on Linux

### LINKED ISSUE #480: writing stuck at 99%
raspberry pi imager 1.7.2 is writing to the sd card and getting stuck at 99%

OS: Manjaro Linux x86_64 
Kernel: 5.19.1-3-MANJARO 

If selecting my own image downloaded from the page It freezes at 99%
However when choosing the OS from the imager interface it works

fkcurrie:
Prevents the '99% hang' issue where the kernel accumulates gigabytes of dirty pages, causing the final fsync() to stall for minutes on slow media.

- Add posix_fadvise(..., POSIX_FADV_DONTNEED) to WriteSequential
- Hints kernel to flush data immediately and free page cache
- Distributes I/O load throughout the write process instead of one massive blocking sync
- Maintains full data integrity (no timeouts)

Fixes #480

### DIFF
--- a/src/linux/file_operations_linux.cpp
+++ b/src/linux/file_operations_linux.cpp
@@ -9,15 +9,39 @@
 #include <unistd.h>
 #include <sys/stat.h>
 #include <errno.h>
+#include <future>
+#include <chrono>
+#include <thread>
 
 #ifndef O_SYNC
 #define O_SYNC O_DSYNC
 #endif
 
-namespace rpi_imager {
+// Timeout for sync operations (in seconds)
+// This prevents indefinite hangs on problematic USB/SD card drivers
+constexpr int SYNC_TIMEOUT_SECONDS = 30;
 
-LinuxFileOperations::LinuxFileOperations() : fd_(-1), last_error_code_(0) {}
-
+/* On Linux, the kernel caches writes in memory (page cache) to improve
+ * performance. When writing large files to slow devices (like SD cards), this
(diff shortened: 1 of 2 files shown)


lurch:
You've got two separate commits in this PR which seem to be interacting in a weird way?

lurch:
You've still got something very wrong here - `FileError LinuxFileOperations::SyncWithTimeout(bool use_fsync) {` is now appearing twice in `file_operations_linux.cpp` :facepalm: 

Also, the behaviour you describe was only for the 1.x series of RPi Imager, and in Imager 2.0 it regularly syncs the data to disk during the writing process; see https://github.com/raspberrypi/rpi-imager/commit/8cdba8759eac2ab6e859d2444f2b1a3ba6c06ca4

tdewey-rpi:
I fear you've slightly misunderstood how the system operates - all platforms determine at the start of the image write a time & data amount interval to perform a flush. This explicitly prevents the scenario you have described - and is trivially monitored in the logs.

On my macBook, for example, each 'batch' is 256MiB, or 7 seconds. It does have the side effect of showing up slow write paths _very_ acutely, but at least you _definitely_ get what you asked for.

tdewey-rpi:
Closing this PR - while well-intentioned, the approach is fundamentally wrong, and doesn't show any sign of having considered the current mechanisms available in the application.

PR #1350, opened 2025-12-01
Romanian translation added

valimaties:
Romanian translation added.
Note: Some UI texts could not be translated because they are not present in the English source file.

### DIFF
--- a/src/i18n/rpi-imager_ro.ts
+++ b/src/i18n/rpi-imager_ro.ts
@@ -1,0 +1,2115 @@
+<?xml version="1.0" encoding="utf-8"?>
+<!DOCTYPE TS>
+<TS version="2.1" language="ro_RO">
+<context>
+    <name>AppOptionsDialog</name>
+    <message>
+        <source>App Options</source>
+        <translation>Opțiuni Aplicație</translation>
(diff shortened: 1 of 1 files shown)


lurch:
> Note: Some UI texts could not be translated because they are not present in the English source file.

Is this still the case? Could you please provide examples of strings which can't be translated?

tdewey-rpi:
I'm simply astounded, @valimaties

You've translated the whole application, it looks like, in one shot. That's incredibly impressive.

Checked with machine translation, happy to add Romanian.

valimaties:
> > Note: Some UI texts could not be translated because they are not present in the English source file.
> 
> Is this still the case? Could you please provide examples of strings which can't be translated?

@lurch , look here:

image

this one is a record from this listview, I cannot find it in english translation, nor in code. Maybe my VS Code did not find it, I don't know, but that text is in english...

image

these ones, also appear in English, even if above there are texts in Romanian.

And there are more...

tdewey-rpi:
All of those strings are hosted by the Imager JSON file - and are not part of the application. There _is_ a mechanism to translate them, however it's not currently used (and is extremely onerous).

PR #1361, opened 2025-12-03
fix(docs): correct typos found during code review

marcelpetrick:
Non-functional changes only:
- Fixed minor spelling mistakes in comments
- Corrected typos in user-facing strings
- No variables, logic, or functional code was modified.

Signed-off-by: Marcel Petrick <mail@marcelpetrick.it>


lurch:
The files in the `dependencies` directory come from 3rd-party projects, so it probably doesn't make sense to fix typos in those?
(Of course if you're feeling super-keen you're welcome to try and get those typos fixed in their upstream projects)

marcelpetrick:
> The files in the `dependencies` directory come from 3rd-party projects, so it probably doesn't make sense to fix typos in those? (Of course if you're feeling super-keen you're welcome to try and get those typos fixed in their upstream projects)

Thank you for pointing this out. You are right and I have reverted those two changes.  
Changing code of dependencies just leads to confusion next time they are updated.

debian/changelog
```diff
@@ -211,10 +211,10 @@ rpi-imager (1.8.2) unstable; urgency=medium
   * Home: Device filtering reworked to download OS list eagerly
   * OS List: Fixed 'Recommended' tag annotation
   * i18n: Added zh-TW Traditional Chinese translation
-  * i18n: Updated Korean, German, Ukranian, translations
+  * i18n: Updated Korean, German, Ukrainian, translations
```
tdewey-rpi: While I appreciate the review - changing previous changelog entries is not something I generally countenance: it feels a little too much like trying to revise history, even if the history you're revising is careless typos.

lurch:
The `os-list-schema.html` file is autogenerated, so you should also fix https://github.com/raspberrypi/rpi-imager/blob/main/doc/json-schema/os-list-schema.json#L398 otherwise your correction will be lost the next time we regenerate `os-list-schema.html` :wink:

PR #1375, opened 2025-12-09
Czech translation added

walken11:
Added file with Czech translation


tdewey-rpi:
Thanks for the submission, @walken11

PR #1381, opened 2025-12-11
Translate all strings into French

martignoni:


tdewey-rpi:
Awesome, thanks @martignoni

PR #1390, opened 2025-12-13
Optimize I/O operations for better performance on MAC

mcyork:
- Remove redundant Flush() call before ForceSync() in periodic sync
- Skip sync operations when using direct I/O (no page cache to flush)
- Optimize macOS writes: use buffered I/O with /dev/rdisk for faster writes
- Enable direct I/O only for verification reads to ensure accuracy

These optimizations improve write performance while maintaining data integrity of verify.

### DIFF
--- a/src/downloadthread.cpp
+++ b/src/downloadthread.cpp
@@ -1471,11 +1471,8 @@ void DownloadThread::_periodicSync()
         return;
     }
     
-    // Skip periodic sync when direct I/O is enabled (F_NOCACHE on macOS, O_DIRECT on Linux).
-    // With direct I/O, data bypasses the page cache entirely, so periodic fsync() provides
-    // no benefit - the data is already going directly to the device. This avoids the severe
-    // performance penalty of frequent fsync() calls on slow media like SD cards over USB.
-    // The final sync at the end of the write is still performed to ensure data integrity.
+    // Skip periodic syncs when using direct I/O - there's no page cache to flush
+    // Direct I/O bypasses the OS page cache, so syncing is unnecessary and wasteful
     if (_file && _file->IsDirectIOEnabled()) {
         return;
     }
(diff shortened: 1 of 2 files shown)


src/mac/file_operations_macos.cpp
```diff
@@ -118,9 +118,18 @@ FileError MacOSFileOperations::OpenDevice(const std::string& path) {
...
     
+    // Convert /dev/diskX to /dev/rdiskX for raw device access (like dd uses)
+    // Raw devices (/dev/rdisk) have better buffering behavior than block devices (/dev/disk)
+    std::string rawPath = path;
+    size_t diskPos = rawPath.find("/dev/disk");
+    if (diskPos != std::string::npos) {
+      rawPath.replace(diskPos + 5, 4, "rdisk");  // Replace "disk" with "rdisk"
+      std::cout << "Using raw device path: " << rawPath << " (instead of " << path << ")" << std::endl;
+    }
+
```
tdewey-rpi: Huh? This is already happening: https://github.com/raspberrypi/rpi-imager/blob/179efafabd1781f93c5e5b1113680109831f3f3a/src/downloadthread.cpp#L323

src/downloadthread.cpp
```diff
@@ -1499,17 +1496,8 @@ void DownloadThread::_periodicSync()
...
-            _writeTimingStats.totalSyncMs.fetch_add(syncMs);
-            _writeTimingStats.syncCount.fetch_add(1);
-            emit eventPeriodicSync(static_cast<quint32>(syncMs), false, currentBytes);
-            qDebug() << "Warning: Flush() failed during periodic sync";
-            return;
-        }
-        
-        // Force filesystem sync using unified FileOperations
+        // ForceSync() already handles both flushing and syncing to disk
+        // No need for redundant Flush() call - ForceSync() does everything
```
tdewey-rpi: The point of this functionality is to avoid excessive page cache usage, which was cited in early testing as causing problems on memory-constrained systems.

tdewey-rpi:
Thanks for the contribution, @mcyork!

I particularly appreciate someone taking the time to model out their ideas in a PR, and *measure* the output! (Honestly, given the scarcity of community-sourced PRs, I nearly fell out of my chair when I saw this!)

That said, I think there's a block you can probably remove - and the Buffered I/O change will need to be considered against earlier testing results (where systems under memory pressure behaved poorly - extremely long 'flush' cycles, allocation thrashing in other applications, etc). Imager 2.0 tries to be very deliberate about it's memory allocation and management - mostly so that if it's causing problems, I can hold my hands up and consider biasing the algorithm further.

tldr; Not an immediate merge, but a valued contribution that I'll evaluate further.

mcyork:
Yeah I wasn't watching the other branches, I just had a write problem so
tried to get my copy going faster. And it worked. Yeah I had to push the PR
just to say clearly what I had done. If you see it in code, it's in code.
You know what it is. Yeah I don't pretend to understand the caching
behavior. I did see that it ran a lot faster. I might have missed another
option where you can increase the buffer. I think there was some line that
said how much to write into the buffer and one of the things that slowed
the Mac down was putting too small a chunk in there and it would use it too
quickly and I think it was waiting for the program to feed it more
chunks. I don't remember but long story short I'm glad you reviewed it and
you seem to be on a stable path. I'm personally not concerned about memory
constraints so if there's a way to say how much memory do I have... then do
this other thing that's nutty :)

Ian McCutcheon
https://workshop.esoup.net/

Issue #1378, opened 2025-12-11
[BUG]: Some minor interface issues


nafanz:
### What happened?

2.0.2: Bugfixes & UI fixes

Russian is used.

1) The "Select your Raspberry Pi device" line is missing; I think the window should automatically adjust to the line size.
<img width="802" height="611" alt="Image" src="https://github.com/user-attachments/assets/1a0574b7-0636-4f68-bab0-653e3fbc318c" />

2) Similar behavior occurs with "When disabling the system disk filter..."; in this case, the "Show system disks" button also extends beyond the borders.
<img width="802" height="611" alt="Image" src="https://github.com/user-attachments/assets/11f0237a-963f-4dc6-bcf4-62503c7d158b" />

3) "Application settings" radio buttons should be right-aligned, I think.
<img width="802" height="611" alt="Image" src="https://github.com/user-attachments/assets/1f3978f0-f836-4076-9396-355b03924af4" />

### Version

2.0.0 (Default)

### What host operating system were you using?

Debian and derivatives (eg Ubuntu)

### Host OS Version

Ubuntu 25.04

### Selected OS

Raspberry Pi OS

### Which Raspberry Pi Device are you using?

Raspberry Pi 5, 500, and Compute Modules 5

### What kind of storage device are you using?

microSD Card in an internal reader

### OS Customisation

- [ ] Yes, I was using OS Customisation when the bug occurred.

### Relevant log output

```shell

```

tdewey-rpi:
Good catches, @nafanz 

I'm slightly reticent to adjust the window size, however - in the past, this ran afoul of concerns for people running on low-resolution displays (read: Netbook-esque hardware). I'll have a think on how to best mitigate the overflows - but any feedback in this sense is welcome.

nafanz:
Yeah, I didn't think about small displays. Just in case, I'll note that I have a Dell U2412M.

nafanz:
I'll point out one more thing about localization.

If I run `rpi-imager` from the command line, which is probably rare, it doesn't work.

The message text has no translation because it's already been installed.

<img width="802" height="611" alt="Image" src="https://github.com/user-attachments/assets/5257be18-19c6-438e-8462-f8736257390c" />

https://github.com/raspberrypi/rpi-imager/blob/1188689422e85170ddbaa6baeb59a7d74b68758f/src/main.cpp#L255

Is that how it should be? Were translations planned for it?

tdewey-rpi:
[...]

PR #1409, opened 2025-12-21
Fix translation of permissions-message dialog

lurch:
The strings were being "translated" too early, so move them to after the QTranslator is set up


lurch:
Fixes part of #1378

PR #1414, opened 2025-12-23
Template default version 2.0.3

nafanz:
For the sake of order


lurch:
@nafanz Looks like you might want to create a local branch for your change and PR from that, rather than PR-ing from your `main` branch? As GitHub now thinks that this PR isn't modifying any files!

nafanz:
https://github.com/raspberrypi/rpi-imager/pull/1435

PR #1480, opened 2026-01-25
feat(installer): Add windows installer dark mode support in rpi-imager.iss.in

rc-chuah:
Add windows installer dark mode support in [rpi-imager.iss.in](https://github.com/raspberrypi/rpi-imager/blob/main/src/windows/rpi-imager.iss.in).


lurch:
Is there a risk that if we add dark-mode support to the Windows installer, people will also then expect Raspberry Pi Imager itself to offer a dark-mode? :thinking:

rc-chuah:
> Is there a risk that if we add dark-mode support to the Windows installer, people will also then expect Raspberry Pi Imager itself to offer a dark-mode? :thinking: 

That’s a really good consideration! In this PR, the dark‑mode support is purely about aligning with Windows installer conventions, so it doesn’t imply theming changes inside Raspberry Pi Imager itself. If people do ask about dark‑mode in Raspberry Pi Imager after seeing this, we can clarify that it’s a separate feature request and decide based on demand. I can also add a note in the PR description to make that distinction clear.

tdewey-rpi:
Thanks for the submission @rc-chuah 

Looks good to me, and good catch!

rc-chuah:
> Thanks for the submission @rc-chuah 
> 
> Looks good to me, and good catch!

You're welcome 😊.

PR #1501, opened 2026-02-03
 Add "Clear saved customisation" button to App Options        

### LINKED ISSUE #1498: [BUG]: can't delete cached metadata
### What happened?

there is no way to just delete all cached customisation and have a clean fresh install again.
the hostname, Localisation, etc cant be deleted
this is a big deal because some of those customisations like the username password can't became blank again, leaving blank means using previously saved input.

### Version

2.0.6 (Default)

macOS

### Host OS Version

Tahoe 26.2

### Selected OS

RPI OS 64

### OS Customisation

### Relevant log output

```shell

```

4RH1T3CT0R7:
- Adds a "Clear" button in App Options to bulk-delete all saved OS customisation settings (hostname, WiFi, user credentials, locale, SSH)                       
  - Includes a confirmation dialog before clearing                                                                                                                
  - Resolves the issue where users had no way to perform a clean fresh install without previously saved settings being silently reused                            
                                                                                                                                                                  
  Fixes #1498                                                                                                                                                     
                                                                                                                                                                  
  Changes                                                                                                                                                         
                                                                                                                                                                  
  - imagewriter.h / imagewriter.cpp — new clearSavedCustomisationSettings() method that removes all keys from the imagecustomization QSettings group              
  - AppOptionsDialog.qml — new ImOptionButton + BaseDialog confirmation, following the existing confirmDisableWarnings pattern (including screen reader and       
  keyboard navigation support)

### DIFF
--- a/src/imagewriter.cpp
+++ b/src/imagewriter.cpp
@@ -3203,6 +3203,14 @@ void ImageWriter::removePersistedCustomisationSetting(const QString &key)
     _settings.sync();
 }
 
+void ImageWriter::clearSavedCustomisationSettings()
+{
+    _settings.beginGroup("imagecustomization");
+    _settings.remove("");
+    _settings.endGroup();
+    _settings.sync();
+}
+
 bool ImageWriter::imageSupportsCustomization()
 {
     return !_initFormat.isEmpty();
--- a/src/imagewriter.h
+++ b/src/imagewriter.h
@@ -283,6 +283,7 @@ public:
     Q_INVOKABLE QVariantMap getSavedCustomisationSettings();
     Q_INVOKABLE void setPersistedCustomisationSetting(const QString &key, const QVariant &value);
     Q_INVOKABLE void removePersistedCustomisationSetting(const QString &key);
+    Q_INVOKABLE void clearSavedCustomisationSettings();
     Q_INVOKABLE bool imageSupportsCustomization();
     Q_INVOKABLE bool imageSupportsCcRpi();
 
--- a/src/wizard/dialogs/AppOptionsDialog.qml
+++ b/src/wizard/dialogs/AppOptionsDialog.qml
@@ -36,7 +36,8 @@ BaseDialog {
         chkEject.naturalWidth,
         chkTelemetry.naturalWidth,
         chkDisableWarnings.naturalWidth,
-        editRepoButton.naturalWidth
+        editRepoButton.naturalWidth,
+        clearSettingsButton.naturalWidth
     ) + Style.cardPadding * 4  // Double padding: contentLayout + optionsLayout margins
     
     // Register focus groups when component is ready
@@ -58,6 +59,7 @@ BaseDialog {
             // Only include secure boot key button if visible
             if (secureBootKeyButton.visible)
                 items.push(secureBootKeyButton.focusItem)
+            items.push(clearSettingsButton.focusItem)
             return items
         }, 1)
         registerFocusGroup("buttons", function(){ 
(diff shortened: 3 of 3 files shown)


tdewey-rpi:
Thanks for the submission, @4RH1T3CT0R7 - I'm particularly impressed that you followed the pattern for screen reader support, and appear to have designed this carefully.

One thing we should consider, though: Users already had the ability to write an OS 'fresh' using the 'skip customisation' button. Do you have an opinion on whether that button is now effectively redundant?

lurch:
It's not something that you've added @4RH1T3CT0R7 , but I've just noticed that our code has an awkward mix of "Customisation" / "Customization" spellings :confused:  E.g. [`getSavedCustomisationSettings`](https://github.com/raspberrypi/rpi-imager/blob/main/src/imagewriter.h#L283) vs. [`imageSupportsCustomization`](https://github.com/raspberrypi/rpi-imager/blob/main/src/imagewriter.h#L286)
@tdewey-rpi Should we prefer one spelling in the code over the other, for consistency? (or am I just being too pedantic again?)

lurch:
> One thing we should consider, though: Users already had the ability to write an OS 'fresh' using the 'skip customisation' button. Do you have an opinion on whether that button is now effectively redundant?

IMHO those two buttons serve different purposes - the "skip customisation" button is still the only way to skip all the customisation questions, and ensure that **no** customisations are applied to the image? Whereas this new button is just "forget my previously-used answers".

tdewey-rpi:
> > One thing we should consider, though: Users already had the ability to write an OS 'fresh' using the 'skip customisation' button. Do you have an opinion on whether that button is now effectively redundant?
> 
> IMHO those two buttons serve different purposes - the "skip customisation" button is still the only way to skip all the customisation questions, and ensure that **no** customisations are applied to the image? Whereas this new button is just "forget my previously-used answers".

That's certainly how I saw it, but we don't have a metrics hook on the button, so I'm soliciting feedback from the author on if the alternative flow is clearer:

Before:
- Press 'skip' at any stage
- No explicit clear-everything button, clear stage-by-stage

After:
- Explicit clear-everything button
- Remove 'skip', to not customise clear-everything

lurch:
@4RH1T3CT0R7 I've just noticed that the commits in this PR include:
* Handle overestimated extract sizes: switch to indeterminate progress for unexpected large writes and cap verification progress at 100%.
* Revert "Handle overestimated extract sizes: switch to indeterminate progress for unexpected large writes and cap verification progress at 100%."

Could you squash the PR down to a single commit please? Thank you.

4RH1T3CT0R7:
@tdewey-rpi @lurch I agree with lurch's take here — the two buttons serve different purposes and should coexist. "Skip customisation" lets the user write an image with no customisations applied at all, while "Clear saved customisation" simply forgets previously saved answers, allowing the user to start fresh if they want to fill in new settings. Removing "Skip" would force users who don't want any customisation to manually clear everything, which is a worse experience. So I'd prefer to keep both buttons

4RH1T3CT0R7:
> @4RH1T3CT0R7 I've just noticed that the commits in this PR include:
> 
> * Handle overestimated extract sizes: switch to indeterminate progress for unexpected large writes and cap verification progress at 100%.
> * Revert "Handle overestimated extract sizes: switch to indeterminate progress for unexpected large writes and cap verification progress at 100%."
> 
> Could you squash the PR down to a single commit please? Thank you.

done

tdewey-rpi:
Tested on macOS - a delightfully unobtrusive change.

Issue #1502, opened 2026-02-03
[BUG]: No files shown when browsing for custom image


thatguywiththekids:
### What happened?

I opened the imager and chose Raspberry Pi 5 series and custom image. The file dialog however shows "no files in this folder", even after waiting for a couple of minutes. It is also not possible to work around it by pasting the image path into the name field: the "open" button remains grey.

### Version

2.0.6 (Default)

### What host operating system were you using?

Other Linux environment

### Host OS Version

Arch Linux with KDE

### Selected OS

Custom

### Which Raspberry Pi Device are you using?

Raspberry Pi 5, 500, and Compute Modules 5

### What kind of storage device are you using?

microSD Card in a USB reader

### OS Customisation

- [ ] Yes, I was using OS Customisation when the bug occurred.

### Relevant log output

```shell

```

lurch:
Are you using the AppImage supplied at https://github.com/raspberrypi/rpi-imager/releases , or a file from the ArchLinux or AUR repos? If it's the latter, then I'll have to close this issue as those are problems beyond our control that ArchLinux will need to fix themselves. (There have been several problems reported with the ArchLinux-supplied versions of Raspberry Pi Imager.)

thatguywiththekids:
I have tested the appimage now. The file selector works fine then, but the problem that a path cannot be used when directly written into the path window remains.

lurch:
Yeah, I think that's just a side-effect of Imager using a custom-written file-browser (so that we get consistent behaviour across different OSes and different Linux distros). "Fixing" this will be a low priority, but we welcome PRs from the community.

In the meantime, if you want to select local images from "full paths" instead of using the file-browser to manually navigate through directories, you can either use the `--cli` option to enable Imager's CLI mode, or use the CLI-specific AppImage, or you can build a custom local JSON file pointing to your images (see https://github.com/raspberrypi/rpi-imager/tree/main/doc/local_json and https://github.com/raspberrypi/rpi-imager/blob/main/doc/schema-notes.md for details).

Issue #1446, opened 2026-01-10
[BUG]: impossible to choose a .img file


kikibelux:
### What happened?

A bug happened! It's impossible for me to choose a .img file 

### Version

2.0.2 (Default)

### What host operating system were you using?

Other Linux environment

### Host OS Version

endeavouros 

### Selected OS

img file

### Which Raspberry Pi Device are you using?

Raspberry Pi 4B, 400, and Compute Modules 4, 4S

### What kind of storage device are you using?

microSD Card in an internal reader

### OS Customisation

- [ ] Yes, I was using OS Customisation when the bug occurred.

### Relevant log output

```shell

```

user0-07161:
i can reproduce this. it's very annoying.

kikibelux:
`dd` il command line is a solution...

lurch:
A quick search tells me that EndeavourOS is based on ArchLinux, and it seems there are known problems with the 3rd-party Arch and AUR packages of Raspberry Pi Imager (see e.g. #1422 and #1433). Please use the AppImage available from https://github.com/raspberrypi/rpi-imager/releases instead.

tdewey-rpi:
Closing - if you're using an unofficial build, your issue reporting path is via the owner of that package.

On Linux, only the AppImage served by downloads.raspberrypi.com or from this GitHub project are supported.

Issue #1433, opened 2025-12-30
[BUG]: QML error?


lavara123:
### What happened?

```
rpi-imager
Not running with elevated privileges - device access may fail
CacheManager initialized with background thread
FileOperations log callback installed
Starting background cache operations
Starting background drive list polling
Failed to register D-Bus object: ""
QQmlApplicationEngine failed to load component
qrc:/qt/qml/RpiImager/main.qml:78:9: Type WizardContainer unavailable
qrc:/qt/qml/RpiImager/wizard/WizardContainer.qml:1036:9: Type WritingStep unavailable
qrc:/qt/qml/RpiImager/wizard/WritingStep.qml:355:17: Non-existent attached object
Stopping background drive list polling
Cleaning up CacheManager
CacheManager destructor: cleaning up background thread
CacheManager destructor: cleanup complete
```

### Version

2.0.2 (Default)

### What host operating system were you using?

Other Linux environment

### Host OS Version

Arch Linux + Hyparland

### Selected OS

pi os lite 64

### Which Raspberry Pi Device are you using?

Raspberry Pi Zero, Zero W, Zero WH

### What kind of storage device are you using?

microSD Card in an internal reader

### OS Customisation

- [ ] Yes, I was using OS Customisation when the bug occurred.

### Relevant log output

```shell

```

lurch:
Are you using the AppImage available from https://github.com/raspberrypi/rpi-imager/releases or some 3rd-party repackaging of Raspberry Pi Imager? If it's the latter, you'll have to contact that 3rd-party for support, as we obviously have no control over what build settings they may or may not have used.

lavara123:
> Are you using the AppImage available from https://github.com/raspberrypi/rpi-imager/releases or some 3rd-party repackaging of Raspberry Pi Imager? If it's the latter, you'll have to contact that 3rd-party for support, as we obviously have no control over what build settings they may or may not have used.

I'm using an original arch repo and I tried to compile and install it, there is the same error

lavara123:
2.0.3
  ```
rpi-imager
Not running with elevated privileges - device access may fail
libcurl initialized globally
CacheManager initialized with background thread
Detected total system memory: 15910 MB on "Linux"
Optimal async queue depth: 512 (memory budget: 3090 MB, block size: 1024 KB, baseline: 512 )
FileOperations log callback installed
Starting background cache operations
Starting background drive list polling
Failed to register D-Bus object: ""
[...]

Issue #1422, opened 2025-12-24
[BUG]: Version 2.0.3 doesn't allow drilling deep enough into the folder where my image is.


MarkBinner:
### What happened?

I installed version **2.0.3** after spotting its release. I tried to use it to write an image stored on my hard disk. The UI would not allow me to go deep enough into the directory tree to select the image I wanted to flash. It stopped one level short of where the file was stored. The image was an uncompressed img file (picoreplayer 11).

### Version

2.0.2 (Default)

### What host operating system were you using?

Debian and derivatives (eg Ubuntu)

### Host OS Version

Mint 21.3

### Selected OS

Custom

### Which Raspberry Pi Device are you using?

Raspberry Pi 4B, 400, and Compute Modules 4, 4S

### What kind of storage device are you using?

microSD Card in a USB reader

### OS Customisation

- [ ] Yes, I was using OS Customisation when the bug occurred.

### Relevant log output

```shell

```

lurch:
Is this a change in behaviour between 2.0.2 and 2.0.3, or does 2.0.2 also have this problem? How many directories "deep" is the directory that you're unable to navigate into?

MarkBinner:
I can't say I'm afraid. I didn't like 2.0 so I've tried nothing between 2.0 and 2.0.3. The path is: /home/markb/Downloads/PiOSes

jr1221:
For me no files or folders are being shown in the image selector.  Version 2.0.2

s0n1q:
For versions: 2.0.2 and 2.0.3-3-gf51b167b can't select custom image file, even if i write full path to file. 

<img width="687" height="480" alt="Image" src="https://github.com/user-attachments/assets/d74f7ff4-7094-418b-8dc7-a232166658f4" />

jr1221:
It seems the native file explorer is never launching via dbus.

s0n1q:
I have next messages in journalctl:
```
qrc:/qt/qml/RpiImager/qmlcomponents/ImFileDialog.qml:474:31: QML Component: Cannot create delegate
rpi-imager[158688]: qrc:/qt/qml/RpiImager/qmlcomponents/ImFileDialog.qml:477:25: Required property fileURL was not initialized
```

lurch:
[...]

PR #1505, opened 2026-02-05
QML: fix property name mismatch in ImFileDialog

wallentx:
### Problem: QML-based file dialog fails due to deprecated model role usage

This change addresses issues reported in:

- https://github.com/raspberrypi/rpi-imager/issues/1422
- https://github.com/raspberrypi/rpi-imager/issues/1433
- https://github.com/raspberrypi/rpi-imager/issues/1446
- https://github.com/raspberrypi/rpi-imager/issues/1502 
  And maybe a few others.

The native XDG Desktop Portal file dialog requires an active user D-Bus session (at least on Arch-based distros). When no session bus is available, `rpi-imager` correctly falls back to its internal QML-based file picker (`ImFileDialog.qml`). This fallback path is expected and intentional.

However, the fallback file dialog was entirely non-functional due to incorrect use of a Qt model role. As a result, the file list rendered empty and the application emitted repeated QML errors such as:

```
NativeFileDialog: No D-Bus session bus available
qml: OSSelectionListView: Restoring scroll - contentY: 0 savedContentY: 138 contentHeight: 999
qml: OSSelectionListView: Restored to 138
qrc:/qt/qml/RpiImager/qmlcomponents/ImFileDialog.qml:670:35: QML Component: Cannot create delegate
qrc:/qt/qml/RpiImager/qmlcomponents/ImFileDialog.qml:673:29: Required property fileURL was not initialized
qrc:/qt/qml/RpiImager/qmlcomponents/ImFileDialog.qml:475:31: QML Component: Cannot create delegate
qrc:/qt/qml/RpiImager/qmlcomponents/ImFileDialog.qml:478:25: Required property fileURL was not initialized
```

### Root cause

~~The failure was caused by a property name mismatch between the data model and the QML delegates:~~

- ~~`FolderListModel` exposes a role named `fileUrl`~~\
  \~\~[~~https://doc.qt.io/qt-6/qml-qt-labs-folderlistmodel-folderlistmodel.html#details~~](https://doc.qt.io/qt-6/qml-qt-labs-folderlistmodel-folderlistmodel.html#details)
- ~~The delegates in `ImFileDialog.qml` defined a required property named `fileURL`~~

~~Because the property was marked as \`\`, this mismatch caused delegate instantiation to fail for every item, resulting in the errors shown above and a completely empty file picker.~~

To be more precise: the `fileURL` role (uppercase) was actually valid in older Qt versions but has been deprecated since Qt 5.15 and appears to have been removed in Qt 6.10.

\* [Qt 6.9 Docs:](https://doc.qt.io/qt-6.9/qml-qt-labs-folderlistmodel-folderlistmodel.html) List both `fileUrl` and `fileURL (deprecated)`.

\* [Qt 6.10 Docs](https://doc.qt.io/qt-6/qml-qt-labs-folderlistmodel-folderlistmodel.html): List only `fileUrl`.

This explains why the AppImage (built with Qt 6.9) still works despite using the old name, while native builds on newer systems (like Arch with Qt 6.10) fail completely. My fix updates the code to use the modern, supported `fileUrl` role, which ensures compatibility with Qt ≥ 5.15 and future Qt versions.

### Fix

Renamed all occurrences of `fileURL` to `fileUrl` in `src/qmlcomponents/ImFileDialog.qml` to match the roles provided by `FolderListModel`.

### Verification

Built and tested locally. The internal QML file picker now functions correctly when the native file dialog is unavailable.

### DIFF
--- a/src/qmlcomponents/ImFileDialog.qml
+++ b/src/qmlcomponents/ImFileDialog.qml
@@ -475,7 +475,7 @@ BaseDialog {
                     delegate: ItemDelegate {
                         required property int index
                         required property string fileName
-                        required property string fileURL
+                        required property url fileUrl
                         
                         width: (ListView.view ? ListView.view.width : 0)
                         text: "📁 " + fileName
@@ -496,7 +496,7 @@ BaseDialog {
                         }
                         onClicked: {
                             subfoldersList.currentIndex = index
-                            dialog.currentFolder = fileURL
+                            dialog.currentFolder = fileUrl
                         }
                     }
                     ScrollBar.vertical: ScrollBar { policy: ScrollBar.AsNeeded; width: Style.scrollBarWidth }
@@ -512,12 +512,12 @@ BaseDialog {
                     }
                     Keys.onEnterPressed: {
                         if (currentIndex >= 0) {
-                            dialog.currentFolder = model.get(currentIndex, "fileURL")
+                            dialog.currentFolder = model.get(currentIndex, "fileUrl")
                         }
                     }
                     Keys.onReturnPressed: {
                         if (currentIndex >= 0) {
-                            dialog.currentFolder = model.get(currentIndex, "fileURL")
+                            dialog.currentFolder = model.get(currentIndex, "fileUrl")
                         }
                     }
                 }
@@ -552,8 +552,8 @@ BaseDialog {
                         currentFileIndex--
                         // Update selection
                         var fileItem = fileColumn.children[currentFileIndex + 1] // +1 because of up entry
-                        if (fileItem && fileItem.fileURL) {
-                            dialog.selectedFile = fileItem.fileURL
+                        if (fileItem && fileItem.fileUrl) {
+                            dialog.selectedFile = fileItem.fileUrl
                         }
                     }
                 }
@@ -562,8 +562,8 @@ BaseDialog {
                         currentFileIndex++
                         // Update selection
                         var fileItem = fileColumn.children[currentFileIndex + 1] // +1 because of up entry
-                        if (fileItem && fileItem.fileURL) {
-                            dialog.selectedFile = fileItem.fileURL
+                        if (fileItem && fileItem.fileUrl) {
+                            dialog.selectedFile = fileItem.fileUrl
                         }
                     }
                 }
@@ -648,7 +648,7 @@ BaseDialog {
                                 currentIndex = 0
                                 // Set the selected file to the first file
                                 if (count > 0) {
-                                    dialog.selectedFile = model.get(0, "fileURL")
+                                    dialog.selectedFile = model.get(0, "fileUrl")
                                 }
                             }
                         }
@@ -657,20 +657,20 @@ BaseDialog {
                         Keys.onUpPressed: {
                             if (currentIndex > 0) {
                                 currentIndex--
-                                dialog.selectedFile = model.get(currentIndex, "fileURL")
+                                dialog.selectedFile = model.get(currentIndex, "fileUrl")
                             }
                         }
                         Keys.onDownPressed: {
                             if (currentIndex < count - 1) {
                                 currentIndex++
-                                dialog.selectedFile = model.get(currentIndex, "fileURL")
+                                dialog.selectedFile = model.get(currentIndex, "fileUrl")
                             }
                         }
                         
                         delegate: ItemDelegate {
                             required property int index
                             required property string fileName
-                            required property string fileURL
+                            required property url fileUrl
                             
                             width: fileColumn.width
                             text: "📄 " + fileName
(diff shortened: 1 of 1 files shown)


lurch:
Interesting... any idea why this only affected the ArchLinux-supplied versions of Raspberry Pi Imager and didn't affect the pre-compiled AppImages that we supply? (see e.g. https://github.com/raspberrypi/rpi-imager/issues/1502#issuecomment-3849787375 )

wallentx:
There are a few things at play here, so bear with me..

## The thing that I'm the most certain of

This is what my PR addresses, and might be best explained by this:
[QT 6.9 Docs](https://doc.qt.io/qt-6.9/qml-qt-labs-folderlistmodel-folderlistmodel.html)
QT6.9 Documentation
[QT 6.10 (latest) Docs](https://doc.qt.io/qt-6/qml-qt-labs-folderlistmodel-folderlistmodel.html)
QT6.10 Documentation

My original PR comment suggesting this being about strictness was incorrect.
In 5.15, they deprecated `fileURL`in favor of `fileUrl`.
In 6.10, they removed `fileURL`.
By convention, the removal of a deprecated feature doesn't get announced when removed, since it is announced at the time of deprecation. It also seems like `qmllint` would have been silent about this.

I don't have a view of your build system, but I pulled the AppImage from the latest release, and was able to determine that it is running `Qt 6.9.3`:
```bash
$ ./Raspberry_Pi_Imager-v2.0.6-desktop-x86_64.AppImage --appimage-extract
$ strings squashfs-root/usr/lib/libQt6Core.so.6 | grep "Qt 6" | head
Qt 6.9.3 (x86_64-little_endian-lp64 shared (dynamic) release build; by GCC 11.4.0)
```

The Arch repo is currently providing `Qt 6.10.2`.

The difference in behavior between versions aligns with what they show in their docs.

## Things I'm less certain about

I believe prior versions of `rpi-imager` used `udisks2` to allow the application to run as a regular user, but I don’t think that’s the case anymore. Since the GUI now makes calls that require elevated permissions directly, the entire GUI itself needs to be run with sudo, which has implications for D-Bus.

I’m _not_ a D-Bus expert, but the default policy on EndeavourOS and Arch uses `auth: EXTERNAL`. _According to Gemini 3 Pro_, this means it “triggers the kernel-level UID check.” I did quite a bit of testing, and neither `sudo`, `sudo -E`, nor `pkexec` was able to access my user’s D-Bus session, which appears to be by design.

As a result, I was unable to get my native build, the Arch repository package, or the AppImage to use my system’s native file picker. All of them fail with `No D-Bus session bus available.` The application does set `XDG*` and `DBUS_SESSION_BUS_ADDRESS` environment variables correctly, but it still fails, so Gemini’s explanation seems plausible to me.

I’d be interested to know whether anyone on another distribution is able to use their system’s native file picker instead of the QML fallback.

lurch:
> > any idea why this only affected the ArchLinux-supplied versions of Raspberry Pi Imager and didn't affect the pre-compiled AppImages that we supply?
> 
> In 5.15, they deprecated `fileURL`in favor of `fileUrl`. In 6.10, they removed `fileURL`. By convention, the removal of a deprecated feature doesn't get announced when removed, since it is announced at the time of deprecation. It also seems like `qmllint` would have been silent about this.
> 
> I don't have a view of your build system, but I pulled the AppImage from the latest release, and was able to determine that it is running `Qt 6.9.3`:
> 
> The Arch repo is currently providing `Qt 6.10.2`.

Ah excellent! Thank you **so** much for your digging on this; "In 6.10, they removed `fileURL`" certainly explains why this only affects the ArchLinux supplied packages.

I'll try to find some time next week to test that AppImages built with your suggested fix included still work okay on Raspberry Pi OS (arm64) and Ubuntu (amd64).

> I believe prior versions of rpi-imager used udisks2 to allow the application to run as a regular user, but I don’t think that’s the case anymore.

Correct, there's more info about that in https://github.com/raspberrypi/rpi-imager/issues/1336#issuecomment-3684293695

But of course as Raspberry Pi Imager is open-source software, the ArchLinux developers are welcome to maintain a downstream patch that re-enables that functionality in the packages that they provide for their distro.

tdewey-rpi:
Good catch, @wallentx - thanks for the submission.

As rpi-imager will migrate to 6.10 fairly soon, this is an extremely valuable catch. Critically, it looks safe to use even in our current vendored 6.9 environment.

tdewey-rpi:
>  believe prior versions of rpi-imager used udisks2 to allow the application to run as a regular user, but I don’t think that’s the case anymore. Since the GUI now makes calls that require elevated permissions directly, the entire GUI itself needs to be run with sudo, which has implications for D-Bus.

That's correct - at least in the current architecture. We observed in internal testing that UDisks2 was completely inoperable in Ubuntu 25.10 - not only did it not prompt, but it threw an auth error back to Imager that presented a UX dead-end. The only viable workaround was to `sudo` the application - which led to this current architectural workaround.

PR #1514, opened 2026-02-15
Fix GenerateTimezones.cmake FALLBACK_FILE path.

uro666:
This fixes the build when built in isolation with no access to the internet.

Prior to this patch `rpi-imager` would fail to find its actual FALLBACK_FILE `timezones.txt` in the source as it was seeking `timezones.txt.fallback`.

This results in builds where users are unable to select nor set their timezone within the `rpi-imager` application and being left with a blank dropdown gui element.

### DIFF
--- a/src/cmake/GenerateTimezones.cmake
+++ b/src/cmake/GenerateTimezones.cmake
@@ -42,7 +42,7 @@ if(download_error)
     message(STATUS "Attempting to use fallback timezone list...")
     
     # Try to find a fallback timezone file
-    set(FALLBACK_FILE "${SOURCE_DIR}/timezones.txt.fallback")
+    set(FALLBACK_FILE "${SOURCE_DIR}/timezones.txt")
     if(EXISTS "${FALLBACK_FILE}")
         message(STATUS "Using fallback timezone file: ${FALLBACK_FILE}")
         file(READ "${FALLBACK_FILE}" fallback_content)
@@ -141,7 +141,7 @@ list(LENGTH timezone_list num_timezones)
 message(STATUS "Generated ${num_timezones} timezone entries in ${OUTPUT_FILE}")
 
 # Compare with fallback file if it exists
-set(FALLBACK_FILE "${SOURCE_DIR}/timezones.txt.fallback")
+set(FALLBACK_FILE "${SOURCE_DIR}/timezones.txt")
 if(EXISTS "${FALLBACK_FILE}")
     message(STATUS "Comparing with existing timezone list...")
     


lurch:
_Perhaps_ the `FALLBACK_FILE` should be set just once (as it is in https://github.com/raspberrypi/rpi-imager/blob/main/src/cmake/GenerateCapitalCities.cmake ), rather than being set in two separate places?

uro666:
> Perhaps the FALLBACK_FILE should be set just once (as it is in https://github.com/raspberrypi/rpi-imager/blob/main/src/cmake/GenerateCapitalCities.cmake ), rather than being set in two separate places?

That would be a sensible idea for a future change to the cmake infra, for now this patch fixes the immediate issue with isolated builds not being able to locate the fallback file at all.

uro666:
It seems that @tdewey-rpi merged my proposed fixes as part of a different patch series without any attribution - https://github.com/raspberrypi/rpi-imager/commit/4dda389d7b612be93624f2cef6b0162ff49eab6c#diff-106b813bcfbaa5a2ef3302894e827980ab37cf5392cf58a9641adf4be626185a

As we are downstream when we as distro package maintainers feed patches back into upstream projects to fix issues with that software as part of being good Linux netizens _in order to provide those fixes for everyone_ by having our submitted PRs merged, receiving the correct attribution for our patches is a small thank you for fixing your software and is simply good manners, ignoring submitted PRs in this way is not conducive to further fixes being sent back upstream.

I have another PR submitted that fixes a Qt Exception when running your software, I would hope it is not treated in the same way as this one - https://github.com/raspberrypi/rpi-imager/pull/1515

Edit to add: upon posting this reply @tdewey-rpi immediately closed the above PR #1515 without any discussion.

lurch:
@uro666 I think there are crossed wires here. Tom didn't "merge your proposed fixes without any attribution", he just happened to independently make the same change, as part of a larger change.
(Apart from maintaining Imager, Tom is also very busy with several other projects)

But as the problem you identified here has now been fixed, I'll close this PR.

uro666:
@lurch my PR sat here for two whole weeks despite you marking the problem it fixes as a bug and Tom making dozens of other commits and PR merges during the same time period, you cannot seriously say he was not aware of this PR existing at all.

Call it crossed wires if that excuse makes you feel better, but from my viewpoint that is not what it looks like at all, especially considering he closed my other PR without so much as any discussion over it at all and that PR fixes a Qt runtime exception with `rpi-imager` itself, it also fixes two separate bug reports that were closed as "Not Planned" despite them showing the same Qt runtime exceptions as that PR addresses!

Perhaps Tom should stop vibe-coding changes with claude LLM and pay more attention to the software and testing of it outwith appImages because he would have found both of the bugs my PR's fix through compiling and running the software locally in an internet isolated environment, as I did.

tdewey-rpi:
> my PR sat here for two whole weeks despite you marking the problem it fixes as a bug and Tom making dozens of other commits and PR merges during the same time period, you cannot seriously say he was not aware of this PR existing at all.

I'm pretty sure he can, given he understands what was going on inside those two weeks and you do not.

I'm extremely unimpressed with your conduct, @uro666 and would caution you against acting in bad faith. Your commentary so far is failing the 'Be Excellent To Each Other' bar, and while I'm not keen to introduce a formal code of conduct, this thread will serve as another chip on the scale in favour of doing so.

Now, to the patch. Imager isn't built offline, and on Linux is released as AppImages only. Complaining that other configurations don't work is akin to complaining to a fish that it can't walk. If you wanted to change that, this wasn't the way to win hearts and minds.

I consider this matter resolved. Further escalation will result in an enforcement action to ensure that community standards are maintained.

PR #1522, opened 2026-02-25
fix showing % >100 for local images

Sundownersport:
Changes made

  src/imagewriter.h — Added _needsDecompressScan flag to the member variable list (line 453)

  src/imagewriter.cpp — 3 changes:
  - Constructor (line 119): Initialize _needsDecompressScan to false
  - setSrc() (line 494): Reset _needsDecompressScan = false on new source selection
  - _parseGzFile() (line 2309): Set _needsDecompressScan = true alongside the existing _extractSizeKnown = false
  - Both LocalFileExtractThread creation sites (lines 828-830 and 3563-3565): Pass the flag via setNeedsDecompressScan()

  src/localfileextractthread.h — Added:
  - setNeedsDecompressScan() public setter (line 23)
  - _scanDecompressedSize() protected method declaration (line 31)
  - _needsDecompressScan private member (line 41)

  src/localfileextractthread.cpp — 2 changes:
  - New _scanDecompressedSize() method (lines 179-217): Uses libarchive to decompress the entire file and count bytes, exactly like your Rust pre-scan. Emits "Scanning compressed image size..." status, resets file position after scan.
  - In run() (lines 84-92): Calls the scan before extractImageRun() when _needsDecompressScan is true, updates _extractTotal with the exact decompressed size.

  The flow is: user selects a .gz file → _parseGzFile() flags it for scanning → worker thread pre-scans the full decompressed size via libarchive → sets accurate extractTotal → extraction/writing begins with
  correct progress tracking.

### DIFF
--- a/src/qmlcomponents/ImComboBox.qml
+++ b/src/qmlcomponents/ImComboBox.qml
@@ -39,248 +39,72 @@ ComboBox {
     // Enhanced properties
     property bool indicateError: false
     
-    // Internal search functionality
+    // Filter-as-you-type state
     property string searchString: ""
-    property real lastKeyTime: 0
-    property int lastSearchIndex: -1
-    property int originalIndex: -1  // Store original selection when popup opens
-    
-    // Keyboard navigation settings - more discoverable timeout
-    readonly property int searchResetTimeout: 1500  // 1.5 seconds - gives users time to type multiple chars
-    readonly property int scrollAnimationDuration: 100  // Smooth animation for scrolling
+    property int originalIndex: -1
+    property var fullModelData: []
     
     // Scroll settings for native-like wheel behavior
     readonly property int itemHeight: 40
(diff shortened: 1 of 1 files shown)


tdewey-rpi:
I explicitly avoided decompressing the entire file just to get the size - the decompression is an I/O expensive operation, and is potentially as expensive as writing the storage media.

What have you done to mitigate this concern?

Sundownersport:
ahhhh dang your right. i'm used to files less than 4gb. Ill redo it with a sample decompressed to extrapolate the file size roughly

tdewey-rpi:
Not a worry, @Sundownersport 

I've made a change for 2.0.7 that will use an indeterminate progress bar with a write counter for files that we can't size a priori. This means you'll still get progress, but you may not know where it'll end - about as good as we can offer without decompressing ahead of time.

PR #1530, opened 2026-02-27
fix(ImFileDialog): allow typing/pasting full file paths in open mode

### LINKED ISSUE #1502: [BUG]: No files shown when browsing for custom image

4RH1T3CT0R7:
- The path field in open mode previously only supported folder navigation. When a user typed a complete file path (e.g. `~/Downloads/image.img`) and pressed Enter, it was incorrectly treated as a directory path and failed silently.
- Add file-path detection via extension heuristic so that typed file paths are accepted directly, and enable the Open button when a file-like path is entered.
- All changes are gated by `!isSaveDialog`, so save mode is unaffected.

Closes #1502

### DIFF
--- a/src/qmlcomponents/ImFileDialog.qml
+++ b/src/qmlcomponents/ImFileDialog.qml
@@ -28,6 +28,7 @@ BaseDialog {
     // Use Dialog's built-in accepted/rejected signals; do not redeclare to avoid overrides
 
     property string _currentFilename: suggestedFilename
+    property string _pathFieldFile: ""
     
     function open() {
         // For save dialogs, initialize filename from suggestion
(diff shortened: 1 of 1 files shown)


tdewey-rpi:
I'm essentially happy with this change. Thanks for the submission, @4RH1T3CT0R7

PR #1535, opened 2026-03-03
fix(beep): bundle fallback chime for Linux distros without freedesktop sounds

### LINKED ISSUE #1449: [BUG]: No Sound with "Play sound when finished"
### What happened?

Play sound when finished option used. Upon finishing, no sound is heard and console says: "Failed to open sound file."

### Version

2.0.3

Other Linux environment

### Host OS Version

Linux

### Selected OS

na

### OS Customisation

### Relevant log output

```shell
Write complete, switching to slow drive scanning
Drive scan mode changed to: Slow
Pipeline timing summary: decompress= 30436 ms (ring_wait= 0 ms)
Write timing stats: writes= 768 syscall= 201264 ms preHashWait= 0 ms postHashWait= 0 ms sync= 0 ms syncCount= 0 avgSize= 8192 KB
PerformanceStats: Cycle ended, success: true events: 21 samples: dl= 159 dec= 639 wr= 768 vfy= 512
Failed to open sound file.
```

4RH1T3CT0R7:
- Expand system sound file search paths to cover Ubuntu (Yaru), KDE (Ocean), and GNOME legacy themes
- Bundle a small (26KB) WAV chime as last-resort fallback for distros with no system sounds at all
- Extract the bundled chime to a PID-namespaced temp file with restrictive permissions (security: rpi-imager runs as root)
- Clean up temp file automatically on application exit via `std::atexit()`

## Root Cause

On Linux distros that don't ship freedesktop sound theme files (e.g. Arch), `findSoundFile()` returns `nullptr` because only two paths were checked:
- `/usr/share/sounds/freedesktop/stereo/complete.oga`
- `/usr/share/sounds/freedesktop/stereo/bell.oga`

This caused `isBeepAvailable()` to return `false` (unless `canberra-gtk-play` or `beep` was installed), and the `beep()` function to fail with "Failed to open sound file." even when `pw-play` or `aplay` was available.

The fix adds three more distro-specific paths, and when none exist, extracts a bundled WAV chime from Qt resources to a temp file. This enables the existing `pw-play`/`aplay`/`pactl` fallback chain to work on any Linux system.

## Security considerations

Since rpi-imager runs as root via pkexec, temp file handling is security-sensitive:
- **PID in filename** prevents symlink attacks and multi-instance collisions
- **`QFile::remove()` before `copy()`** prevents following pre-existing symlinks
- **`ReadOwner | WriteOwner` permissions** restricts access to root only
- **`std::atexit()` cleanup** removes the temp file on exit

Closes #1449


src/linux/platformquirks_linux.cpp
```diff
@@ -326,23 +327,52 @@ void applyQuirks() {
...
+                }
+            }
+
+            // 2. Extract bundled fallback chime to a temp file.
+            //    Use PID in filename to avoid symlink attacks and multi-instance collisions
+            //    (rpi-imager runs as root via pkexec, so temp file security matters).
+            QFile bundled(":/sounds/chime.wav");
+            if (bundled.exists()) {
+                static QString tempPath = QDir::tempPath()
+                    + QString("/rpi-imager-chime-%1.wav").arg(QCoreApplication::applicationPid());
```
tdewey-rpi: I'd be happier if it used a random source - predicting pid allocation feels fairly straightforward.

tdewey-rpi:
What's the copyright story on the chime.wav? Are you able to certify it's your own product and you're able to assign usage rights to the project?

4RH1T3CT0R7:
Good catch, thanks! I initially assumed the chime was from @0xDE57's repo (https://github.com/0xDE57/raspberry-chime, CC0 1.0), but after your comment I double-checked and it turned out to be a different file entirely

Replaced it now with `314A.wav` from that repo - properly CC0-licensed, 88KB, root note at 314Hz (a nice pi reference)

tdewey-rpi:
@0xDE57

Confirming we're now in a place to accept this. Thanks for the contributions both!

PR #1539, opened 2026-03-07
fix(progress): clamp write and verify progress to 100%

### LINKED ISSUE #1491: [BUG]: "Writing..." is exceeding 100%
### What happened?

A bug happened!

Image

248 % until going to "Verification" step, which didn't exceed 100%

imaging completed without errors

### Version

2.0.6 (Default)

### Host OS Version

Debian 13.3 KDE Plasma

### Selected OS

Batocera

microSD Card in a USB reader

### OS Customisation

### Relevant log output

```shell

```

4RH1T3CT0R7:
Clamp the calculated write and verify progress percentage to [0, 100] so the progress bar never exceeds 100%

The async write callback in `DownloadThread` captures `_extractTotal` at the moment each write is queued. If the total is stale or underestimated (e.g. for custom OS images like Batocera), `bytesWritten` can exceed the total, producing percentages like 248%

For `.gz` files >4GB this is already handled via indeterminate progress mode, but other formats with inaccurate size estimates were not covered

## Fix

Add `Math.min(..., 100)` to the progress calculation in both `onWriteProgress` and `onVerifyProgress` handlers in `WritingStep.qml`

Closes #1491

### DIFF
--- a/src/wizard/WritingStep.qml
+++ b/src/wizard/WritingStep.qml
@@ -589,7 +589,7 @@ WizardStepBase {
                 var bytesWrittenMB = Math.round(now / (1024 * 1024))
                 progressText.text = qsTr("Writing... %1 MB written").arg(bytesWrittenMB)
             } else {
-                var progress = total > 0 ? (now / total) * 100 : 0
+                var progress = total > 0 ? Math.min((now / total) * 100, 100) : 0
                 progressBar.value = progress
                 progressText.text = qsTr("Writing... %1%").arg(Math.round(progress))
             }
@@ -601,7 +601,7 @@ WizardStepBase {
             root.isVerifying = true
             root.bottleneckStatus = ""  // Clear write bottleneck during verification
             root.operationWarning = ""  // Clear write warnings during verification
-            var progress = total > 0 ? (now / total) * 100 : 0
+            var progress = total > 0 ? Math.min((now / total) * 100, 100) : 0
             progressBar.value = progress
             progressText.text = qsTr("Verifying... %1%").arg(Math.round(progress))
         }


tdewey-rpi:
> Clamp the calculated write and verify progress percentage to [0, 100] so the progress bar never exceeds 100%
> 
> The async write callback in `DownloadThread` captures `_extractTotal` at the moment each write is queued. If the total is stale or underestimated (e.g. for custom OS images like Batocera), `bytesWritten` can exceed the total, producing percentages like 248%
> 
> For `.gz` files >4GB this is already handled via indeterminate progress mode, but other formats with inaccurate size estimates were not covered
> 
> ## Fix
> Add `Math.min(..., 100)` to the progress calculation in both `onWriteProgress` and `onVerifyProgress` handlers in `WritingStep.qml`
> 
> Closes #1491

What other formats are estimating size?

4RH1T3CT0R7:
Two scenarios where `bytesWritten` can exceed the captured `totalBytes`:

1. **Gzip files >4GB** - the ISIZE field in the gzip footer is 32-bit, so it wraps around. The code already handles this with `isIndeterminateProgress` (showing "X MB written" instead of %), but the `Math.min` clamp acts as a safety net in case that detection is ever bypassed

2. **Remote compressed downloads without `extract_size` in the manifest** - when the manifest doesn't provide `extract_size`, `_extrLen` stays 0 and `setExtractTotal` falls back to `_downloadLen` (the compressed size). For compressed formats the actual extracted size can be much larger than `_downloadLen`, so `bytesWritten` exceeds the total

The `Math.min` clamp is purely defensive - it doesn't change any size estimation logic, just ensures the UI never displays >100% regardless of the source

tdewey-rpi:
> Two scenarios where `bytesWritten` can exceed the captured `totalBytes`:
> 
> 1. **Gzip files >4GB** - the ISIZE field in the gzip footer is 32-bit, so it wraps around. The code already handles this with `isIndeterminateProgress` (showing "X MB written" instead of %), but the `Math.min` clamp acts as a safety net in case that detection is ever bypassed
> 2. **Remote compressed downloads without `extract_size` in the manifest** - when the manifest doesn't provide `extract_size`, `_extrLen` stays 0 and `setExtractTotal` falls back to `_downloadLen` (the compressed size). For compressed formats the actual extracted size can be much larger than `_downloadLen`, so `bytesWritten` exceeds the total
> 
> The `Math.min` clamp is purely defensive - it doesn't change any size estimation logic, just ensures the UI never displays >100% regardless of the source

Thanks for the reply, @4RH1T3CT0R7 

I'm not inclined to take this patch on two grounds:

1. We already filter out gzip files for this reason, and this patch would hide any path that circumvented the existing signalling behind a harder-to-detect error (a long dwell at 100%, rather than an obvious miscalculation)
2. extract_size is a required manifest field, and this would again potentially hide a more serious failure (see https://github.com/raspberrypi/rpi-imager/blob/1aef52f9b06e0226bd52e6de07a5f47e7dfe1432/doc/schema-notes.md?plain=1#L173 for reference)

While I appreciate the defensive approach, in this case I would prefer to fail-fast (and obviously).

4RH1T3CT0R7:
Makes sense - fail-fast with an obvious miscalculation is better than silently dwelling at 100%. Thanks for the explanation on the `extract_size` requirement; I missed that it's mandatory in the schema

Closing this one

Issue #1520, opened 2026-02-21
[BUG]: Enabling USB Gadget Mode persists between sessions resulting it being applied without displaying the option


phattmatt:
### What happened?

Hi,

I noticed when writing recent images that the USB Gadget Mode was being applied, even though I wasn't using a custom manifest, and the option was not presented to me during the customisation process.

I believe that the setting has been saved to the registry during an image write while using a custom minifest, and is being applied in subsequent sessions, even though the default online manifest doesn't include the "usb_otg" capability.

To disable the USB Gadget Mode, I can load the custom manifest, which means the GUI shows the option, disable the USB Gadget Mode option, and then quit. Re-running RPI-Imager results in the USB Gadget Mode not being applied.

### Version

2.0.6 (Default)

### What host operating system were you using?

Windows

### Host OS Version

Windows 11 25H2

### Selected OS

Raspberry Pi OS Lite (64-bit)

### Which Raspberry Pi Device are you using?

Raspberry Pi 5, 500, and Compute Modules 5

### What kind of storage device are you using?

microSD Card in a USB reader

### OS Customisation

- [x] Yes, I was using OS Customisation when the bug occurred.

### Relevant log output

```shell
Copy of my Raspberry Pi Imager Registry Entry which shows the persisted setting:

Windows Registry Editor Version 5.00

[HKEY_CURRENT_USER\Software\Raspberry Pi\Raspberry Pi Imager\imagecustomization]
"capitalCity"="London (United Kingdom)"
"enable1Wire"="false"
"enableI2C"="false"
"enableSPI"="false"
"enableSerial"="Disabled"
"enableUsbGadget"="true"
"hostname"="rpi5"
"keyboard"="gb"
"recommendedWifiCountry"="GB"
"sshEnabled"="true"
"sshPasswordAuth"="true"
"sshUserName"="pi"
"sshUserPassword"="Redacted"
"timezone"="Europe/London"
"wifiHidden"="false"
"wifiMode"="secure"
"wifiPasswordCrypt"="Redacted"
"wifiSSID"="Redacted"
```

tdewey-rpi:
> I believe that the setting has been saved to the registry during an image write while using a custom minifest, and is being applied in subsequent sessions, even though the default online manifest doesn't include the "usb_otg" capability.

That's a good catch! The customisation flags got persisted into QSettings, and it looks like that included the experimental flags.

tdewey-rpi:
[...]

PR #1540, opened 2026-03-07
fix(wifi): prevent stale WiFi settings from being applied after Skip

### LINKED ISSUE #1326: [BUG]: wifi builds default wifi on
### What happened?

When I attempted a headless build working on a wifi attached unit, configuring for ETH0 only, wifi settings were forced.

### Version

2.0.0 (Default)

### Host OS Version

RPi OS Trixie

### Selected OS

RPi OS Trixie Light

### OS Customisation

### Relevant log output

```shell

```

4RH1T3CT0R7:
When a user previously configured WiFi and later skips customization (e.g. for an ETH0-only headless build), the WiFi settings from the previous session were silently applied to the image

`customizationSettings` is loaded from persistent storage (QSettings) at app startup. When the user clicks "Skip" on any customization step, the runtime flags (`wifiConfigured = false`) are cleared but the WiFi keys (`wifiSSID`, `wifiPasswordCrypt`, `wifiHidden`) remain in `customizationSettings`. The customization generator checks `wifiSSID` directly - not the `wifiConfigured` flag - so it applies the stale WiFi config

This is the same class of bug as #1520 (USB Gadget Mode persistence)

## Fix

Before calling `applyCustomisationFromSettings()`, strip WiFi keys from `customizationSettings` when `wifiConfigured` is `false`. This is done centrally in `WizardContainer.qml` at the writing step entry point, covering all Skip paths

Closes #1326

### DIFF
--- a/src/customization_generator.cpp
+++ b/src/customization_generator.cpp
@@ -72,9 +72,10 @@ QByteArray CustomisationGenerator::generateSystemdScript(const QVariantMap& s, c
     const QString userPass = s.value("sshUserPassword").toString(); // crypted if present
     const QString sshPublicKey = s.value("sshPublicKey").toString().trimmed();
     const QString sshAuthorizedKeys = s.value("sshAuthorizedKeys").toString().trimmed();
-    const QString ssid = s.value("wifiSSID").toString();
-    const QString cryptedPskFromSettings = s.value("wifiPasswordCrypt").toString();
-    bool hidden = s.value("wifiHidden").toBool();
+    const bool wifiConfigured = s.value("wifiConfigured", true).toBool();
+    const QString ssid = wifiConfigured ? s.value("wifiSSID").toString() : QString();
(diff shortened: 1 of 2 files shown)


tdewey-rpi:
Thanks for the PR, @4RH1T3CT0R7 

I think this is probably inverted, though? Shouldn't we just fix the customisation generator to actually honour the `wifiConfigured` flag, and avoid otherwise transforming state?

4RH1T3CT0R7:
Good point, the generator should honour the flag directly rather than filtering state in QML

Reworked: the `wifiConfigured` flag is now passed through to `customization_generator.cpp`, and all three generation paths (`generateSystemdScript`, `generateCloudInitUserData`, `generateCloudInitNetworkConfig`) check it. When `wifiConfigured` is `false`, WiFi-related fields (`ssid`, `cryptedPsk`, `hidden`) are set to empty/false, so stale persisted values are never used. Defaults to `true` for backward compatibility

The QML side now just passes `customizationSettings.wifiConfigured = wifiConfigured` and hands the complete settings object to the generator unchanged

PR #1555, opened 2026-03-24
fix(i18n): add missing French translations for saved customisation settings

Ferdinaelectro1:
Some strings added recently in AppOptionsDialog.qml for the "Saved Customisation" feature were missing from the French translation file.


tdewey-rpi:
Thanks for the submission, @Ferdinaelectro1 

I have confirmed with machine translation that this contains nothing that obviously breaches accepted norms, so I'm happy to accept this patch for release in 2.0.7.

Issue #1466, opened 2026-01-19
[FEATURE]: Option to run without root privileges


mikeri:
### Is your feature request related to a problem? Please describe.

Root is not needed in all cases, and applications insisting on requiring root when they don't really need it is extremely frustrating. Now we have to run it with root privileges even we they are not needed.

### Describe the solution you would like to see implemented

Instead of the message about needing root with exit at  the only option, we should be able to run the imager without root permission.

### Describe alternatives you've considered

Instead of the current option that only allow gaining privileges or exiting, I suggest an option with something like "OK, I understand" to allow users to run the imager without root privileges.

### Additional context

Power users have udev rules to allow writing to usb storage without sudo, for several reasons: one, we never risk writing to system disks by accident when using cat or dd to write images to (micro)SD cards. (For example ```cat rpios.img > /dev/sdd``` without sudo works when /dev/sdd is a usb drive). Second, we don't need to run as many apps as root.

### Version

2.0.4 (Default)

lurch:
Closing as a duplicate of several other issues, including #1336 , #1351 and #1354

Issue #1336, opened 2025-11-29
[BUG]: cannot run appimage as non-root user on debian ( wayland )


flo-dhalluin:
### What happened?

downloaded the new 2.0.0 app image from official raspberry pi site, on startup I have big modal telling me the image needs root permission, then only choice is exit. 
I can't run graphical app as root ( because wayland and appimages), I know my user has the right perms to flash sd cards because previous rpi-imager versions worked fine. 
Is there a way to bypass this permission check ?

### Version

2.0.0 (Default)

### What host operating system were you using?

Windows

### Host OS Version

debian testing

### Selected OS

raspberrypi os lite 64

### Which Raspberry Pi Device are you using?

Raspberry Pi 4B, 400, and Compute Modules 4, 4S

### What kind of storage device are you using?

microSD Card in a USB reader

### OS Customisation

- [ ] Yes, I was using OS Customisation when the bug occurred.

### Relevant log output

```shell

```

Shikakiben:
appimages have no problem to run in wayland in this case its a Qt problem on wayland.

run in terminal:

sudo env QT_QPA_PLATFORM=xcb /path/to/rpi-imager

if it doesn't work try:

xhost +SI:localuser:root
sudo env QT_QPA_PLATFORM=xcb /path/to/rpi-imager
xhost -SI:localuser:root

EDIT: QT_QPA_PLATFORM=xcb doesn't seem to be needed

xhost +SI:localuser:root
sudo  /path/to/rpi-imager
xhost -SI:localuser:root

tdewey-rpi:
You must run as root to access the block devices you want to write to. We can no longer rely on the mechanisms employed by previous Imager versions.

As such, a mechanism to bypass that requirement would be counterproductive.

Closing as 'Won't fix', because Imager is behaving as expected.

septatrix:
> You must run as root to access the block devices you want to write to.

For writing yes, but the UI does not need to and should never run as root. On one hand does that open many security issues (e.g. downloading and such all run as root, and vulnerabilities in the network stack are very dangerous) and on the other hand does this mean the app cannot run on Wayland and instead must run via XWayland

> We can no longer rely on the mechanisms employed by previous Imager versions.

Why not? What mechanism was previously used?

I am not sure how this is handled on Windows etc but on Linux it is definitely possible to elevate privileges using polkit (e.g. with systemd-run) to only run some parts as root.

tdewey-rpi:
[...]

PR #1557, opened 2026-03-25
feat(linux): Allow running without root

### LINKED ISSUE #1354: [BUG]: Imager refuses to run without root
### What happened?

It looks like the imager checks if it's running as root, and refuses to launch otherwise, claiming it won't be able to write to storage devices otherwise.

While I get that it's easier to just tell people to run the imager with sudo, I much prefer chowning the storage medium instead. It's possible to write to storage media without absolute root privileges, therefore I think it's bad design to flatout refuse to run without them.

### Version

2.0.0 (Default)

Other Linux environment

### Host OS Version

Arch

### Selected OS

N/A

### OS Customisation

### Relevant log output

```shell
ERROR: Not running as root.
Writing to storage devices requires elevated privileges.
Please run with sudo: sudo Raspberry_Pi_Imager-v2.0.0-cli-x86_64.AppImage --cli ...
exit 1
```

darkwater:
Closes #1354 

An additional "continue anyway" button in the dialog might be nice too, but at least this now provides the possibility of running it without root.

### DIFF
--- a/src/Style.qml
+++ b/src/Style.qml
@@ -89,6 +89,7 @@ Item {
     /// the check-boxes/radio-buttons have labels that might be disabled
     readonly property color formLabelColor: "black"
     readonly property color formLabelErrorColor: "red"
+    readonly property color formLabelWarningColor: "#e8a000"
     readonly property color formLabelDisabledColor: "grey"
     // Active color for radio buttons, checkboxes, and switches
     readonly property color formControlActiveColor: "#1955AE"
--- a/src/cli.cpp
+++ b/src/cli.cpp
@@ -58,6 +58,7 @@ int Cli::run()
         {"quiet", "Only write to console on error"},
         {"log-file", "Log output to file (for debugging)", "path", ""},
         {"secure-boot-key", "Path to RSA private key (PEM format) for secure boot signing", "key-file", ""},
+        {"non-root", "Allow running without elevated privileges"}
     });
 
     parser.addPositionalArgument("src", "Image file/URL");
@@ -65,7 +66,7 @@ int Cli::run()
     parser.process(*_app);
 
     // Check for elevated privileges on platforms that require them (Linux/Windows)
-    if (!PlatformQuirks::hasElevatedPrivileges())
+    if (!PlatformQuirks::hasElevatedPrivileges() && !parser.isSet("non-root"))
     {
         // Common error message
         const char* commonMsg = "Writing to storage devices requires elevated privileges.";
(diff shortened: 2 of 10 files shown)


tdewey-rpi:
Thanks for the submission, @darkwater.

I'm not inclined to accept this.

Where I have explained why Imager behaves in this manner, the primary point is that failing early is better than failing later. Simply adding a way to circumvent the early failure, but without engaging the later failure, is not likely to result in a net confusion reduction. If anything, I could see an enterprising repackager put this flag into their .desktop file, and then create untold confusion in their user base.

To run without root, the application _must_ be fractured into privileged and unprivileged parts.

darkwater:
I had Copilot add some UI to indicate unwritable devices. Would something like this suffice?

image

Running GUI applications as root is generally not a great thing. In fact, Wayland compositors normally don't allow connections from other users, even from root. If I do `sudo rpi-imager` right now, it simply doesn't work. I understand wanting to avoid confusion, but this cannot come at the cost of functionality.

tdewey-rpi:
> I understand wanting to avoid confusion, but this cannot come at the cost of functionality.

I couldn't disagree more. Functionality that users cannot meaningfully drive is pointless, with history and art littered with examples:

TheHomer

I'll circle back around to this PR at some stage, but I note it's got some fairly extensive merge conflicts.

lurch:
I _suspect_ that this PR is targetting the same thing as #1466 i.e. "Power users have udev rules to allow writing to usb storage without sudo", and IMHO that's _such_ a small proportion of our total user-base that the additional friction this would add for the majority of our users outweighs the benefit it would provide to a handful of Linux power users. See my earlier comment [here](https://github.com/raspberrypi/rpi-imager/issues/1336#issuecomment-3773136515).

darkwater:
> I couldn't disagree more. Functionality that users cannot meaningfully drive is pointless, with history and art littered with examples:

I honestly don't understand your logic. Should I reply with an image from some animated movie where every corner has been smoothed off and butter knifes are illegal because they could hurt someone? Just because not everyone understands an option, doesn't mean it shouldn't exist. What beginner user would even run `rpi-imager --help` on the command line to discover this option without having the knowledge of how to use it?

For any Arch users coming across this issue, I published a `rpi-imager-git-non-root` package to the AUR that patches out the root check. It's clear that my views do not align with this project.

lurch:
You're totally fine to maintain your own downstream fork of Raspberry Pi Imager (that's the beauty of Open Source!) but please don't get upset if the view of the upstream maintainers doesn't match yours.
Raspberry Pi Imager has a very large user-base, and so we have to focus our development efforts on ease of use (don't forget that Raspberry Pi is popular with schoolchildren) and maintainability, rather than keeping a handful of advanced Arch Linux users happy. The more features that get added to Imager, the more effort it takes for us to test each Imager version before release, and the harder it becomes for us to add new features without worrying that they might accidentally break existing seldom-used features.

PR #1559, opened 2026-03-27
Screenshot

nafanz:
A small improvement that allows you to get acquainted with the application interface before installing it and delving into the documentation


tdewey-rpi:
That's....a delightful addition, @nafanz. Happy to accept it.

PR #1570, opened 2026-03-30
Updated translation dictionaries

nafanz:
I fixed some Qt Linguist comments, mostly about extra dots at the end of the translation string that weren't in the original string. I did this for convenience, so I could quickly see in the Qt Linguist interface what really needs attention and avoid getting distracted by these little details.

I noticed that the translation dictionaries weren't containing many strings. I think I managed to update them, but I'm not sure I did everything correctly. I couldn't have done it without Copilot.

```
sudo apt install --no-install-recommends build-essential cmake git libgnutls28-dev libudev-dev pkg-config
sudo ./qt/build-qt.sh
export Qt6_ROOT="/opt/Qt/6.9.3/gcc_64"
export PATH="/opt/Qt/6.9.3/gcc_64/bin:$PATH"
cd src/
mkdir build
cd build
cmake ../src
cmake --build . --target rpi-imager_lupdate
```

If everything is OK, please merge them.

I'll try to do the Russian translations later (I checked quite a few new strings there), unless someone else beats me to it.

### DIFF
--- a/src/i18n/rpi-imager_bn.ts
+++ b/src/i18n/rpi-imager_bn.ts
@@ -115,6 +115,42 @@
         <source>All Files (*)</source>
         <translation type="unfinished"></translation>
     </message>
+    <message>
+        <source>Audio notification unavailable - no viable audio player found on this system</source>
+        <translation type="unfinished"></translation>
+    </message>
+    <message>
+        <source>Saved Customisation</source>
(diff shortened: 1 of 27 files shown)


tdewey-rpi:
I manually re-generated these, @nafanz as other translations were updated in the interim. Thanks for sending this PR through.

PR #1579, opened 2026-04-06
Ignore drives with .rpiignore

andydvsn:
Really simple request; I'd like the Imager to never show me external drives which have an `.rpiignore` file at the root.

The reason is that I have an external drive permanently attached to my Mac, and every time I use the Imager I'm paranoid I'll zap this drive by mistake. It hasn't happened yet! But a simple check for this empty file on the root would make me a lot happier, and I think others may find it helpful too.

### DIFF
--- a/src/drivelistmodel.cpp
+++ b/src/drivelistmodel.cpp
@@ -8,6 +8,7 @@
 #include "drivelist/drivelist.h"
 #include <QSet>
 #include <QDebug>
+#include <QFile>
 
 DriveListModel::DriveListModel(QObject *parent)
     : QAbstractListModel(parent)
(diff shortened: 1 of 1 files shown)


lurch:
IIRC this `processDriveList` function gets called repeatedly while Imager is open, so that it can check for removable drives being added or removed. So if this code is constantly checking for this file of every drive, I guess that would prevent any of the drives connected to the system from going to sleep while Imager is open? (Assuming of course that the OS doesn't cache that information)

tdewey-rpi:
While this isn't an unreasonable change on the face of it, there is a performance consideration to bear in mind on Linux systems that may have _many_ mounted devices (in particular, I note AppImages, Snaps & Flatpaks could lead to this scenario).

As such, I can't merge this until I've observed the performance impact. It may be we need to cache the mountpoint file existence records - or possibly approach this by introducing a UI element and performing an inverted search.

lurch:
And of course this also assumes that the partition containing the `.rpiignore` file remains mounted at all times. If the partition was ever unmounted without the drive being unplugged (which I guess is less likely on macOS or Windows than it is on Linux?), then Imager would happily display the drive again.

andydvsn:
All very good points. I should never have used the word "simple". ;)

lurch:
@andydvsn I guess you might be able to partially-placate Tom's performance concerns by moving your check to later in the function, e.g. after the `if (i.isVirtual)` checks? That way all the "light-weight" checks get run first, and your "heavy-weight" check only gets run if all the light-weight checks have already passed.

tdewey-rpi:
Closing - this has gone somewhat stale, and the design considerations have been understood by all parties. Happy to see another patch later.

PR #1599, opened 2026-04-20
fix: prefer kreadconfig6 with fallback to kreadconfig5

Ferdinaelectro1:
Previously hardcoded kreadconfig5, which fails on KDE Plasma 6.
Now tries kreadconfig6 first, falls back to kreadconfig5 if not found.

### DIFF
--- a/src/linux/platformquirks_linux.cpp
+++ b/src/linux/platformquirks_linux.cpp
@@ -1299,7 +1299,8 @@ bool prefersReducedMotion() {
 
     // KDE Plasma 5.67+: AnimationDurationFactor=0 disables animations
     {
-        const QString bin = QStringLiteral("/usr/bin/kreadconfig5");
+        const QString sessionVersion = qEnvironmentVariable("KDE_SESSION_VERSION");
+        const QString bin = QStringLiteral("/usr/bin/kreadconfig") + sessionVersion;
         if (QFileInfo::exists(bin)) {
             QProcess kreadconfig;
             kreadconfig.start(bin, {"--group", "KDE", "--key", "AnimationDurationFactor"});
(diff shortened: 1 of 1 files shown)


src/linux/platformquirks_linux.cpp
```diff
@@ -1299,7 +1299,8 @@ bool prefersReducedMotion() {
 
     // KDE Plasma 5.67+: AnimationDurationFactor=0 disables animations
     {
-        const QString bin = QStringLiteral("/usr/bin/kreadconfig5");
+        const QString sessionVersion = qEnvironmentVariable("KDE_SESSION_VERSION");
+        const QString bin = QStringLiteral("/usr/bin/kreadconfig") + sessionVersion;
```
lurch: Should this fallback to `/usr/bin/kreadconfig5` if the `KDE_SESSION_VERSION` env-var doesn't exist?
Also, should we validate the `KDE_SESSION_VERSION` env-var against a known regex? (Note the comment at the start of this function which says:
```
    // Use absolute paths for external commands — this function may run in an
    // elevated (root) context after pkexec.  Relative names would search PATH,
    // which could be user-controlled.
```
)

Ferdinaelectro1:
@lurch  Updated in the latest commit, please take another look!

lurch:
> @lurch Updated in the latest commit, please take another look!

Cool. I guess you might want to update the PR description? :wink:

Ferdinaelectro1:
@lurch Done!

lurch:
Would you mind updating the PR title too? Thank you.

Ferdinaelectro1:
@lurch Done ! Thanks

tdewey-rpi (approved):

PR #1655, opened 2026-07-03
Create port and build target for FreeBSD systems

jimc1827:
Ports rpi-imager for FreeBSD systems (tested on the two tier-1 architectures, amd64 and arm64), including instructions and a bourne shell script capable of building both GUI and CLI-only versions.

- Opts to keep logic similar to Linux code where reasonable
- Implement drive list enumeration with posix libraries and FreeBSD's native `CAM` and `GEOM` APIs instead of `lsblk --json`
- Implement asynchronous IO using posix `aio` rather than `IO_uring`
- Prefer to link libraries dynamically by default to conform to FreeBSD ports tree best practices
    - Appimages are not provided due to FreeBSD's lack of support
    - Use libraries packaged into FreeBSD base where possible
    - Statically link `zstd` since it provides experimental functions that are used in `imagewriter.cpp`

Also fixes a potential SEGV in `downloadextractthread.cpp` caused by throwing `runtime_error(nullptr)` in certain cases, which yields UB. This is handled gracefully on Linux, but leads to a segmentation fault on FreeBSD's libc implementation.

A FreeBSD package/port is under review on [FreeBSD's Bugzilla](https://bugs.freebsd.org/bugzilla/show_bug.cgi?id=296480). Whenever this process is complete, a separate commit will be posted updating the README.md to reflect the option to install a binary version of rpi-imager.

Messy development history is largely stored in the repository's [experimental](https://github.com/jainnimish/rpi-imager/tree/experimental) branch and the [devel](https://github.com/jainnimish/rpi-imager/tree/devel) branch, modulo some bugs we found during the commit refactoring process.


src/cmake/GenerateVersion.cmake
```diff
@@ -2,15 +2,16 @@
...
-#   SOURCE_DIR       — repo root (for git describe)
-#   OUTPUT_DIR       — directory for generated files
+#   GIT_EXECUTABLE        — path to git
+#   SOURCE_DIR            — repo root (for git describe)
+#   OUTPUT_DIR            — directory for generated files
+#   DEFAULT_VERSION_STR   — default version string for when git describe fails
 
-find_package(Git QUIET)
-
-set(VERSION_STR "0.0.0-unknown")
```
tdewey-rpi, line 11: ?

jimc1827: To be accepted into FreeBSD Ports, I was told that git should not be used at build time.

I reasoned that if CMakeLists.txt could not find Git, then it is unlikely that GenerateVersion.cmake would do so; therefore, I could move the call to `find_package` into the if statement without much consequence. This has the advantage that I can disable git just once in CMakeLists.txt to prevent it from ever searching for the package.

I allowed the builder to pass the environment variable `DEFAULT_VERSION_STRING`, which defaults to "0.0.0-unknown". This means that in the case where I expect `git describe` to fail (e.g. a FreeBSD ports build), I can manually set a correct version string.

tdewey-rpi:
This is a large change, and one that I'll have to consider deeply. The essential issues are:

1) I don't run FreeBSD - nor do I have bandwidth to add a machine/VM running FreeBSD to my testing regime.
2) Without running FreeBSD, I can't see how to sign off the release of this build
3) As it's not using the same dependency tree as the AppImage builds (or indeed, any other platform), this expands the support cone - again, I don't have bandwidth for that.

So while I greatly appreciate the effort involved, adopting the PR in good faith is challenging at best without an ongoing commitment from a FreeBSD maintainer.

jainnimish:
Hello! Thanks for taking time to reply.

The last few commits have focused on reducing the patchset size to make this PR easier to review.

To this end, have directly imported linux code when we can instead of deduplicating it.  For `bootimgcreator` and `FileOperations`, we created a new `unix` directory with common methods. Platform-specific implementations inherit from this base class. Doing all of this has reduced our patch size by approximately 2k lines. What remains is largely new, FreeBSD-specific code we wrote.

> adopting the PR in good faith is challenging at best without an ongoing commitment from a FreeBSD maintainer.

I think we have found a maintainer. There should be a comment from them soon.

src/cmake/GenerateVersion.cmake
emaste: > To be accepted into FreeBSD Ports, I was told that git should not be used at build time.

Also in FreeBSD ports we'd generally want dependencies to come from other ports, not be built-in or fetched by the upstream project's build infrastructure. If this is at odds with the upstream project's own approach though we'd just have to carry it as a downstream patch.

src/drivelist/drivelist_freebsd.cpp
```diff
@@ -0,0 +1,39 @@
+/*
+ * SPDX-License-Identifier: Apache-2.0
+ * Copyright (C) 2020-2025 Raspberry Pi Ltd
```
emaste, line 3: We'll need to check on the approach wrt copyright assignment

src/unix/PlatformPackaging.cmake
```diff
@@ -0,0 +1,20 @@
+# Common commands used by both FreeBSD and Linux
```
emaste, line 1: (With @tdewey-rpi's confirmation) we should rearrange these commits to pull out the common code first (linux/PlatformPackaging.cmake to unix/PlatformPackaging.cmake), and then use it from the new FreeBSD implementation, rather than copying first followed by deduplication.

tdewey-rpi: Happy with that plan, and with FreeBSD getting it's own implementation (which is my preferred route from a maintenance PoV anyway!)

tdewey-rpi:
> Hello! Thanks for taking time to reply.
> 
> The last few commits have focused on reducing the patchset size to make this PR easier to review.
> 
> To this end, have directly imported linux code when we can instead of deduplicating it. For `bootimgcreator` and `FileOperations`, we created a new `unix` directory with common methods. Platform-specific implementations inherit from this base class. Doing all of this has reduced our patch size by approximately 2k lines. What remains is largely new, FreeBSD-specific code we wrote.
> 
> > adopting the PR in good faith is challenging at best without an ongoing commitment from a FreeBSD maintainer.
> 
> I think we have found a maintainer. There should be a comment from them soon.

Apologies for the delay in getting back to you.

Yes, to be clear: I'd greatly prefer you have a separate PAL implementation for FreeBSD - potentially inheriting from a 'common' PAL backend. The explicit design objective of the PAL components is to avoid having constructs like..

#ifdef Q_OS_WIN

littered through the codebase - which make tracing behaviour an absolute minefield.

Also happy to hear that there's a FreeBSD maintainer lined up. I'll defer further commentary until they signal they're ready to take on the mantle.

PR #1660, opened 2026-07-09
Fix window icon setting for Linux

City-busz:
The ico file format cannot be read on Linux. Use the installed icon from the theme instead.

### DIFF
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -238,7 +238,11 @@ int main(int argc, char *argv[])
     app.setOrganizationDomain("raspberrypi.com");
     app.setApplicationName("Raspberry Pi Imager");
     app.setApplicationVersion(ImageWriter::staticVersion());
+#ifdef Q_OS_WIN
     app.setWindowIcon(QIcon(":/icons/rpi-imager.ico"));
+#else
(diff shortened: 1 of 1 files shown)


Ferdinaelectro1:
@City-busz @tdewey-rpi I noticed a potential issue on Linux with this approach, especially regarding standalone packages like AppImage or when running the compiled binary directly from the build directory.

In these scenarios, the application is not installed globally, meaning the icon is not registered in standard system directories like /usr/share/icons or ~/.local/share/icons. As a result, ` QIcon::fromTheme("rpi-imager")`  will fail and return a null/empty icon, leaving the main window without any visual identity.

To ensure a robust cross-platform behavior, we should implement a fallback mechanism on Linux:

    1- Try to load the icon from the system theme first (to respect native desktop environments and packaging like .deb).

    2- Fall back to the compiled Qt Resource (`:/icons/rpi-imager.ico` ) if the theme icon is null.

tdewey-rpi (changes requested):
While this change is certainly expedient - you've absolutely identified a more involved problem.

As @Ferdinaelectro1 pointed out, different execution environments will ask for different things - and rather than using Q_OS_FOO to swap in different parameters, I think this is better served by pushing this content resolution into the PAL

PR #1674, opened 2026-07-27
2.0.11-rc2: Bugfixes

tdewey-rpi:


dutypro:
Any idea when we can expect this to be pushed?

tdewey-rpi:
> Any idea when we can expect this to be pushed?

There's one further fix that needs to land - a release process change that should make releasing future Raspberry Pi Imagers much simpler compared to the previous flows. I'm actively testing this, so once I've gained confidence, you should expect a -rc2 build.

Issue #1666, opened 2026-07-19
[BUG]: v2.0.11-rc1 fails to unmount drive the first time; always succeeds the second


sastraxi:
### What happened?

I can reliably reproduce that v2.0.11-rc1 will fail to write an image saying it was unable to mount the drive. Immediately afterwards, I'm able to click "Write" again and this time it works. I have reproduced this 3 times in the software. The same image flashes the first time on v2.0.10.

### Version

2.0.11-rc1

### What host operating system were you using?

macOS

### Host OS Version

15.7.2 (24G325)

### Selected OS

Custom (https://github.com/TreeFallSound/pi-gen-pistomp/releases/tag/release%2F3.2.1-rc1)

### Which Raspberry Pi Device are you using?

Raspberry Pi 5, 500, and Compute Modules 5

### What kind of storage device are you using?

microSD Card in the internal macbook reader (M1 Pro).

### OS Customisation

I reproduced this bug both with and without OS customization.

tdewey-rpi:
Fixed in v2.0.11+ full release.

PR #1683, opened 2026-08-06
fix(macos): actually wait for the DiskArbitration unmount callback

elibosley:
On current `main`, every write on macOS begins with an error dialog:

> Failed to unmount disk '/dev/diskN'. Please close any applications using the disk and try again.

The disk is not busy. Dismissing the dialog and writing again succeeds, because the unmount we stopped waiting for had already completed. Reproduced on macOS with two different USB flash drives.

## Cause

8b8695ae ("fix(mac): run DiskArbitration on the main run loop") replaced the blocking wait in `runDiskOperation()` with:

```c
CFRunLoopRunInMode(kCFRunLoopDefaultMode, 0.0, true);
```

A zero-second `CFRunLoopRunInMode()` only drains sources that are *already* pending and returns immediately. The loop is bounded by `maxRetries * 100` = 1200 iterations, so the timeout documented just above it as 60 seconds now elapses in microseconds — far too soon for `diskarbitrationd` to answer. `context.completed` is still false, `runDiskOperation()` returns `DiskResult::Busy`, and `DownloadThread::_openAndPrepareDevice()` surfaces that as the dialog above.

The same commit also wrapped the call in `dispatch_sync(dispatch_get_main_queue(), ...)`. That is invisible today only because the wait returns instantly; restoring a real wait underneath it would freeze the GUI for the duration of every unmount.

## Fix

- Restore the blocking `CFRunLoopRunInMode(..., 0.05, false)` wait, and the `CFRunLoopStop()` calls in the three DiskArbitration callbacks that wake it early, so the retry budget really is the 60 seconds it claims.
- Run the session on the calling thread's run loop rather than dispatching to the main queue. The callers are the write and format worker threads, so blocking there cannot stall the UI.
- Keep the correct half of 8b8695ae: schedule the session before initiating the operation.

Net effect is the pre-8b8695ae wait behaviour, with that commit's scheduling-order fix retained.

## Testing

Syntax-checked the DiskArbitration path with clang. I have not been able to run a full Qt build locally — a macOS write test on CI or by a maintainer would confirm the dialog is gone.

Found while rebasing a downstream fork onto 2.0.11; the affected file is identical to upstream, so the change applies here unmodified.

### DIFF
--- a/src/mac/platformquirks_macos.mm
+++ b/src/mac/platformquirks_macos.mm
@@ -112,20 +112,23 @@ namespace {
         DiskOpContext* ctx = static_cast<DiskOpContext*>(context);
         ctx->result = translateDissenter(dissenter);
         ctx->completed = true;
+        CFRunLoopStop(CFRunLoopGetCurrent());
     }
-    
+
     void ejectCallback(DADiskRef disk, DADissenterRef dissenter, void* context) {
         (void)disk;
         DiskOpContext* ctx = static_cast<DiskOpContext*>(context);
         ctx->result = translateDissenter(dissenter);
         ctx->completed = true;
+        CFRunLoopStop(CFRunLoopGetCurrent());
     }
-    
+
     void ejectUnmountCallback(DADiskRef disk, DADissenterRef dissenter, void* context) {
         DiskOpContext* ctx = static_cast<DiskOpContext*>(context);
         if (dissenter) {
             ctx->result = translateDissenter(dissenter);
             ctx->completed = true;
+            CFRunLoopStop(CFRunLoopGetCurrent());
         } else {
             // Unmount succeeded, now eject
             DADiskEject(disk, kDADiskEjectOptionDefault, ejectCallback, context);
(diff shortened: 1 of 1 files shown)


elibosley:
@tdewey-rpi this started after pulling the latest .11

elibosley:
Closing this — I hadn't spotted that you'd already fixed it in #1674 (`07e2f57c`, plus the `c92f01f6` refactor on top). Same root cause, and your version is the better shape: keeping the session on the main run loop and blocking the worker on a semaphore avoids the nested-run-loop question entirely, and tearing the session down inside a `dispatch_sync` to the main queue serialises teardown against callback delivery, which closes a stack-lifetime hole my patch left open.

Sorry for the duplicate noise. Flagging it only because it's biting real users today: we ship a downstream fork and every macOS write in our 2.0.11 build opened with "Failed to unmount disk", so it may be worth landing that pair ahead of the rest of #1674 if the release-process change keeps that PR waiting. Happy either way — closing this in favour of yours.

tdewey-rpi:
> Sorry for the duplicate noise. Flagging it only because it's biting real users today: we ship a downstream fork and every macOS write in our 2.0.11 build opened with "Failed to unmount disk", so it may be worth landing that pair ahead of the rest of #1674 if the release-process change keeps that PR waiting. Happy either way — closing this in favour of yours.

Not a problem - it's deeply encouraging when members of the community find the same bugs!

I'm expecting the 2.0.11 release to emerge today - final checks on the new release infrastructure are underway.

sastraxi:
Sorry to necro, just confirming this is the same bug?
https://github.com/raspberrypi/rpi-imager/issues/1666

PR #1685, opened 2026-08-08
feat(gui): run the post-write eject in the background with live status

elibosley:
## Problem

After a write finishes, the eject runs synchronously inside the write/extract threads. On macOS the unmount inside `ejectDisk()` is what flushes the freshly written data out of the page cache, so it can run for tens of seconds on a slow stick — measured with the DiskArbitration code against hdiutil FAT32 volumes: **6.5 s** to eject an empty 64 MB volume, **10.1 s** after writing 250 MB across 1000 files. A slow physical stick with a large multi-file payload can exceed the 60-second timeout entirely.

During that window the UI either freezes on "Finalising…" or shows the done screen already claiming *"The storage device was ejected automatically. You can now remove it safely."* — while the kernel is still writing to the drive. Users who trust that text can pull a stick mid-flush and corrupt it; users who don't get what looks like a hung app.

Today this is partially masked on `main` because the zero-second `CFRunLoopRunInMode()` polling loop (from 8b8695ae) makes the eject return in microseconds without waiting — but that also means the "safe to remove" text has never actually been true. Once DiskArbitration callbacks are delivered again (#1674 fixes this), the synchronous eject genuinely blocks and the freeze becomes user-visible. We shipped this exact sequence in our downstream fork (unraid/usb-creator-next): fixing the DA wait immediately produced "app hangs after write, force-quit needed" reports, which this restructure fixed.

## Change

Two commits:

**`fix(macos)`** — raise the eject's DiskArbitration timeout to five minutes (unmount keeps 60 s). The callers treat the result as best-effort, but timing out early lets the UI claim safety mid-flush. This composes with either DA wait implementation, including #1674's.

**`feat(gui)`** — never let user-visible state get ahead of the hardware:

- `DownloadThread` gains `ejectStarted()` / `ejectFinished(bool)` and a shared `_performEject()`; every write path announces the eject, emits `success()` immediately, then ejects on its worker thread — the done screen appears the moment the write finishes.
- `ImageWriter` exposes an `EjectState` property (`Idle / InProgress / Succeeded / Failed`) plus `Q_INVOKABLE ejectDrive()` for a manual eject/retry on a disposable thread. The thread is never `terminate()`d (a thread blocked in a DA `mach_msg` may never receive the deferred cancellation, after which deleting the still-running `QThread` is a qFatal); teardown waits briefly and otherwise leaves it to finish detached behind a `QPointer` guard. Retry is protected against the queued result arriving before `QThread::finished` (pointer-match cleanup).
- The done screen shows the live status — *"Ejecting the storage device — do not remove it yet…"* → *"…ejected. You can now remove it safely."*, or a failure message with an **Eject** button — and disables Finish / Write Another while an eject is in flight. The safe-removal text now only appears after an eject actually succeeded, and the eject text/button are hidden for fastboot targets, which have no removable medium.

## Testing

- All touched C++ TUs syntax-checked against Qt 6; QML linted.
- The equivalent change shipped downstream in unraid/usb-creator-next v2.0.11-unraid.2-rc.2 (PR unraid/usb-creator-next#121, CodeRabbit-reviewed) and is in QA on real hardware there now.
- Needs a maintainer-side macOS write to see the full effect once #1674 restores real DA waits; on current `main` the UI changes are visible but the eject still returns instantly.

## Relationship to #1674

Complementary, not conflicting: #1674 fixes *how* the DA callbacks are delivered; this PR fixes *when* the eject runs relative to `success()` and what the UI claims meanwhile. The only shared file is `platformquirks_macos.mm`, where this PR's change is the one-line eject timeout (plus comment) — trivial to rebase whichever lands second.

### DIFF
--- a/src/imagewriter.cpp
+++ b/src/imagewriter.cpp
@@ -2696,12 +2696,7 @@ void ImageWriter::ejectDrive()
     QThread *thread = QThread::create([self, device]() {
         PlatformQuirks::DiskResult result =
             PlatformQuirks::ejectDisk(PlatformQuirks::getEjectDevicePath(device));
-        bool succeeded = (result == PlatformQuirks::DiskResult::Success);
-#ifdef Q_OS_WIN
-        // The legacy Windows result can carry an unrelated volume's failure
-        // even when the target drive ejected fine; only report a definite miss.
-        succeeded = (result != PlatformQuirks::DiskResult::InvalidDrive);
-#endif
+        const bool succeeded = (result == PlatformQuirks::DiskResult::Success);
         QMetaObject::invokeMethod(QCoreApplication::instance(), [self, succeeded]() {
             if (self)
                 self->onEjectFinished(succeeded);
--- a/src/imagewriter.cpp
+++ b/src/imagewriter.cpp
@@ -46,6 +46,7 @@
 #include <QTimeZone>
 #include <QNetworkInterface>
 #include <QCoreApplication>
+#include <QPointer>
 #ifndef CLI_ONLY_BUILD
 #include <QQmlContext>
 #include <QWindow>
@@ -509,6 +510,22 @@ ImageWriter::~ImageWriter()
         qDebug() << "Cleaning up CacheManager";
         delete _cacheManager;
         _cacheManager = nullptr;
+    }
+
+    // Give a user-triggered eject a chance to finish before teardown. Never
+    // terminate() it: killing the thread mid-DiskArbitration call is unsafe,
+    // and deleting a QThread that survived terminate() is a qFatal. If it is
+    // still flushing after the grace period, leave it to finish detached — the
+    // worker only reaches this object through a guarded QPointer, so it cannot
+    // touch a destroyed ImageWriter.
+    if (_manualEjectThread) {
+        qDebug() << "Waiting for manual eject thread to finish";
+        if (_manualEjectThread->wait(10000)) {
+            delete _manualEjectThread;
+        } else {
+            qWarning() << "Manual eject still running at teardown; leaving it to finish detached";
+        }
+        _manualEjectThread = nullptr;
     }
 
     // Ensure any running thread is properly cleaned up
@@ -1028,6 +1045,7 @@ void ImageWriter::startWrite()
     }
 
     setWriteState(WriteState::Preparing);
+    setEjectState(EjectState::EjectIdle);
 
     if (_isFastbootDevice)
     {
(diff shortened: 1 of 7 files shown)


src/wizard/DoneStep.qml
```diff
@@ -234,10 +237,22 @@ WizardStepBase {
             }
             FocusableText {
                 id: ejectInstruction
-                text: root.autoEjectEnabled ? qsTr("The storage device was ejected automatically. You can now remove it safely.") : qsTr("Please eject the storage device before removing it from your computer.")
+                visible: root.ejectApplicable
+                text: {
+                    if (root.ejectState === ImageWriterSingleton.EjectInProgress)
+                        return qsTr("Ejecting the storage device — do not remove it yet…")
+                    if (root.ejectState === ImageWriterSingleton.EjectSucceeded)
+                        return qsTr("The storage device was ejected. You can now remove it safely.")
```
tdewey-rpi, line 245: Please retain 'ejected automatically' - this will preserve the translation at the cost of one additional word.

src/imagewriter.cpp
```diff
@@ -2620,6 +2640,72 @@ void ImageWriter::restartWrite(QString reason)
...
+    // destroyed ImageWriter.
+    QPointer<ImageWriter> self(this);
+    QThread *thread = QThread::create([self, device]() {
+        PlatformQuirks::DiskResult result =
+            PlatformQuirks::ejectDisk(PlatformQuirks::getEjectDevicePath(device));
+        bool succeeded = (result == PlatformQuirks::DiskResult::Success);
+#ifdef Q_OS_WIN
+        // The legacy Windows result can carry an unrelated volume's failure
+        // even when the target drive ejected fine; only report a definite miss.
+        succeeded = (result != PlatformQuirks::DiskResult::InvalidDrive);
```
tdewey-rpi, line 2692: When I asked claude to review this patch, it suggested that this check was too lenient - and it would mask busy/error results. My Mk1 eyeball agrees. Please could you remove the Q_OS_WIN block here, and then we can just rely on the DiskResult that's returned?

tdewey-rpi (changes requested):

elibosley:
Thanks for the review! I pushed both changes, and I rebased the branch onto `main` at v2.0.11.1.

**`src/wizard/DoneStep.qml`**

Restored. The `EjectSucceeded` case now uses the original string:

> The storage device was ejected automatically. You can now remove it safely.

**`src/imagewriter.cpp`**

Removed. The result is now `succeeded = (result == PlatformQuirks::DiskResult::Success)`.

I removed the same block in `DownloadThread::_performEject()`.

PR #1705, opened 2026-08-18
fix(linux): wait for the partition node before the post-format mount

elibosley:
The multi-file flow formats the target and then mounts the first partition to extract into. The manual-mount fallback ran `mount(8)` immediately, but the node for a freshly written partition table appears asynchronously: the kernel re-reads the table when the last writer closes the whole-disk device, and the exclusive open added in `b70f9083` can make that re-read fail with `EBUSY` and happen later still.

We hit this in CI as:

```
Cross-platform disk formatter succeeded in 25 ms
mount: /dev/sda1: special device does not exist
```

passing on rerun — so it is a race, not a hard failure.

This waits up to ten seconds for the node, and asks the kernel for a `BLKRRPART` re-read if it has not appeared after two. When the node is already present the wait costs one `stat`. A cancel during the wait returns instead of falling through to mount.

The automount poll above needs no equivalent: it discovers mountpoints the OS created, and its Linux fallback is this path. Customisation reaches the FAT filesystem through whole-disk offsets and never touches the partition node.

Not compiled here (no Qt toolchain on this machine); CI covers the build.

### DIFF
--- a/src/downloadextractthread.cpp
+++ b/src/downloadextractthread.cpp
@@ -29,6 +29,13 @@
 #include <unistd.h>
 #endif
 
+#ifdef Q_OS_LINUX
+#include <sys/ioctl.h>
+#include <linux/fs.h>   // BLKRRPART
+#include <cerrno>
+#include <QFile>
+#endif
+
 using namespace std;
 
 // Ring buffer slot count is now determined dynamically by SystemMemoryManager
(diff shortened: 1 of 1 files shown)


src/downloadextractthread.cpp
```diff
@@ -29,6 +29,13 @@
 #include <unistd.h>
 #endif
 
+#ifdef Q_OS_LINUX
+#include <sys/ioctl.h>
+#include <linux/fs.h>   // BLKRRPART
+#include <cerrno>
+#include <QFile>
+#endif
+
```
tdewey-rpi, line 38: Not a blocker for this PR - but as a heads-up:

I'm increasingly trying to push anything that would've been Q_OS_FOO into the PAL layer, to try and improve readability - and flag to other OS impls where special handling may be required. This is particularly relevant for the upcoming FreeBSD maintainer, whose early PR is working through review.

tdewey-rpi (approved):

Issue #1532, opened 2026-03-01
[BUG]: Scroll direction for language and time zone are wrong on Windows


Benjin:
### What happened?

My OS is set to regular scroll: scrolling down on the touchpad scrolls down, not touchscreen-style scrolling where scrolling down on the touchpad scrolls up.

While the Raspberry Pi OS selection scrolls as expected, the dropdowns in the locale page (timezone, keyboard layout, capitol city) are reversed.

### Version

2.0.6 (Default)

### What host operating system were you using?

Windows

### Host OS Version

Windows 11 25H2

### Selected OS

64 bit latest

### Which Raspberry Pi Device are you using?

Raspberry Pi 3B, 3A+, 3B+, and Compute Module 3, 3+

### What kind of storage device are you using?

microSD Card in an internal reader

### OS Customisation

- [ ] Yes, I was using OS Customisation when the bug occurred.

### Relevant log output

```shell

```

PR #1713, opened 2026-08-24
fix(ImComboBox): scroll the dropdown in the direction of the wheel

### LINKED ISSUE #1589: [BUG]: Selection boxes in localization scroll opposite from mouse wheel
### What happened?

Using the bleeding edge, `3ca0078964bcd23f2f408a1637dee36f9ab01e83`, when making selections for localization, the selection boxes scroll opposite from the direction the mouse wheel is turned.

### Version

2.0.8 (Default)

### Host OS Version

Devuan 6

### Selected OS

Raspberry Pi OS Lite (64-bit)

### OS Customisation

### Relevant log output

```shell

```

4RH1T3CT0R7:
The popup's `WheelHandler` does `contentY + dy`. By that point `dy` has already been through the inversion handling, so a positive value means an "up" gesture, meaning a lower `contentY`. Adding it scrolled the popup backwards unless your desktop is set to natural scrolling. Subtracting fixes it.

The closed-state handler already gets this right, so the same gesture moved the selection opposite ways with the popup open and closed. #1534 fixed the closed half for #1532; this is the other one.

I left `isScrollInverted()` alone, so the popup follows the gesture rather than the OS content direction, same as the closed state. On a natural-scrolling desktop the dropdown now moves with the wheel, not the content. If you'd rather it track the OS direction, drop the inversion here instead and I'll send that version.

Fixes #1589

### DIFF
--- a/src/qmlcomponents/ImComboBox.qml
+++ b/src/qmlcomponents/ImComboBox.qml
@@ -391,8 +391,9 @@ ComboBox {
                     if (PlatformHelper.isScrollInverted(event.inverted)) {
                         dy = -dy
                     }
-                    
-                    var newY = dropdownList.contentY + dy
+
+                    // dy is now positive for an "up" gesture, and up means earlier in the list, so contentY decreases.
+                    var newY = dropdownList.contentY - dy
                     var maxY = Math.max(0, dropdownList.contentHeight - dropdownList.height)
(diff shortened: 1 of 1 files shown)


4RH1T3CT0R7:
@tdewey-rpi this is the other half of #1534, the popup direction. Happy to switch it to follow the OS scroll setting instead if you'd prefer that.

PR #1725, opened 2026-09-03
Keep the write failure dialog open until dismissed

### LINKED ISSUE #1693: [BUG]: Failure message disappears upon activating Imager's window - 2.0.11 rc1
### What happened?

I used Iamger 2.0.11 rc1 ro write an Lite OS to 8GB microSD, worked on other tasks while waiting for Imager to complete, and saw a message appearing on the partially burried Imager window.

Now clicking on the Imager window in order to activate it and see the failure message, the message disappeared. And nothing was left on the window telling me what the message was about! (Can't remember the top header's content)

This is close to a bug. The Imager should show the result until I actively confirm by closing the message.

Windows 10

### Version

2.0.10 (Default)

Windows

### Host OS Version

22H2

### Selected OS

RPi OS Lite 32bit

microSD Card in a USB reader

### OS Customisation

### Relevant log output

```shell

```

4RH1T3CT0R7:
BaseDialog never set a closePolicy, so every dialog built on it inherited the Popup default, which also closes on a press outside. If a write fails while the window is buried, the click that brings it forward dismisses the error dialog unread. The writing step also hides its progress section once the state is Failed, so no trace of the failure remains.

BaseDialog now uses CloseOnEscape only. 2d089f7d already did this for every popup in 2024 for the same reason, and the Qt6 dialog rewrite (beb26ff1) dropped it, hence the app-wide change rather than errorDialog-only. The writing step keeps the "Write failed" line visible until the next write starts. Checked against the write state machine and with qmllint; not run in a build.

Fixes #1693

### DIFF
--- a/src/qmlcomponents/BaseDialog.qml
+++ b/src/qmlcomponents/BaseDialog.qml
@@ -21,6 +21,11 @@ Dialog {
     
     // Standard dialog properties
     modal: true
+
+    // Dismiss only via Escape or the dialog's own buttons. The Popup default also
+    // closes on a press outside, so the click that re-activates a buried window
+    // would silently discard the message (#1693).
+    closePolicy: Popup.CloseOnEscape
     
     // Reset Dialog's built-in padding - we use our own margins in contentLayout
     padding: 0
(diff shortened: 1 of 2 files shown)


tdewey-rpi:
@4RH1T3CT0R7 I wound up using a slightly different variant of this change, and have tagged you in the merged commit. Good steering.

Issue #1727, opened 2026-09-03
[BUG]: `create_local_json.py` fails when creating a local manifest file with local files containing `+`


phattmatt:
### What happened?

Hi,

create_local_json.py fails when a matching local image filename contains a character that is URL-encoded by pathlib.Path.as_uri(), such as +.

For example, with:

ubuntu-24.04.4-preinstalled-server-arm64+raspi.img.xz

the local file is converted to a URI containing:

ubuntu-24.04.4-preinstalled-server-arm64%2Braspi.img.xz

Later, the script sorts the results by comparing os.path.basename(x["url"]) against os_order, which contains the original unencoded filename. This causes:

ValueError: 'ubuntu-24.04.4-preinstalled-server-arm64%2Braspi.img.xz' is not in list

The issue appears to be in the final sorting step. A possible fix is to sort using the existing local_os_entries dictionary keys rather than deriving the filename again from the encoded URI.

Steps to reproduce the error:

```
pi@rpi5nvme:~/Downloads/local_rpi_images $ pwd
/home/pi/Downloads/local_rpi_images
```
```
pi@rpi5nvme:~/Downloads/local_rpi_images $ curl -s https://downloads.raspberrypi.com/os_list_imagingutility_v4.json | jq -r '
  .os_list[]
  | recurse(.subitems[]?)
  | select(.init_format? == "cloudinit")
  | select(.name? | test("^Ubuntu Server (24\\.04|26\\.04)"))
  | .url
' | xargs wget
--2026-09-03 21:31:46--  http://cdimage.ubuntu.com/releases/resolute/release/ubuntu-26.04.1-preinstalled-server-arm64+raspi.img.xz
Resolving cdimage.ubuntu.com (cdimage.ubuntu.com)... 185.125.190.40, 91.189.91.124, 185.125.190.37, ...
Connecting to cdimage.ubuntu.com (cdimage.ubuntu.com)|185.125.190.40|:80... connected.
HTTP request sent, awaiting response... 301 Moved Permanently
Location: http://cdimage.ubuntu.com/ubuntu/releases/resolute/release/ubuntu-26.04.1-preinstalled-server-arm64+raspi.img.xz [following]
--2026-09-03 21:31:46--  http://cdimage.ubuntu.com/ubuntu/releases/resolute/release/ubuntu-26.04.1-preinstalled-server-arm64+raspi.img.xz
Reusing existing connection to cdimage.ubuntu.com:80.
[...]

PR #1728, opened 2026-09-04
Fix local manifest sorting for URL-encoded filenames

phattmatt:
### Summary

Fixes `create_local_json.py` failing when creating a local manifest for image filenames containing characters that are URL-encoded by `Path.as_uri()`, such as `+`.

For example, Ubuntu Server images use filenames such as:

`ubuntu-24.04.4-preinstalled-server-arm64+raspi.img.xz`

When converted to a local `file://` URI, the `+` becomes `%2B`. The existing sorting code then extracts the basename from this encoded URI and tries to find it in `os_order`, which contains the original unencoded filename, resulting in a `ValueError`.

### Fix

Instead of reconstructing the filename from the rewritten local URL, build the output list by iterating over the existing `os_order` filename keys and selecting matching entries from `local_os_entries`.

This preserves the same ordering as the source manifest while avoiding any dependency on URL encoding.

### Tested

Tested using locally downloaded Ubuntu Server 24.04.4 and 26.04.1 Raspberry Pi images. The original code fails with the `%2B`/`+` mismatch; with this change the local manifest is generated successfully and retains the original manifest ordering.

Addresses issue: https://github.com/raspberrypi/rpi-imager/issues/1727

### DIFF
--- a/doc/local_json/create_local_json.py
+++ b/doc/local_json/create_local_json.py
@@ -237,7 +237,11 @@ if __name__ == "__main__":
                     device["capabilities"] = sorted(merged_caps)
     # Sort the output OSes into the same order as the input OSes
     os_order = list(online_os_entries.keys())
-    local_data["os_list"] = sorted(local_os_entries.values(), key=lambda x: os_order.index(os.path.basename(x["url"])))
+    local_data["os_list"] = [
+        local_os_entries[name]
+        for name in os_order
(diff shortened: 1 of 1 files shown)


tdewey-rpi (approved):

tdewey-rpi:
Good catch, thanks @phattmatt

PR #1729, opened 2026-09-04
fix(linux): restore XDG_DATA_DIRS under sudo/pkexec for xdg-open

### LINKED ISSUE #1447: [BUG]: "What is Raspberry Pi Connect?" URL doesn't open anything
### What happened?

Clicking on the "What is Raspberry Pi Connect?" URL in Customization: Raspberry Pi Connect screen does nothing.

"runuser: cannot open session: Module is unknown"

### Version

2.0.3

Other Linux environment

### Host OS Version

Linux

### Selected OS

na

### OS Customisation

### Relevant log output

```shell
Opening URL: "https://www.raspberrypi.com/software/connect/"
Started runuser xdg-open
runuser: cannot open session: Module is unknown
```

4RH1T3CT0R7:
Under pkexec or sudo the wrapper strips XDG_DATA_DIRS, and the snap and flatpak directories only get added to it by login-shell profile scripts. xdg-open and GIO then cannot find the browser's .desktop file and fall back to the text/html handler, so the links open in a text editor. The original report's runuser PAM error went away with 2.0.5; the 2.0.7 reports are this.

applyQuirks() already rebuilds the invoking user's HOME, XDG_*_HOME, runtime dir and bus address. This adds XDG_DATA_DIRS to that set when it is absent, spec default plus the snap and flatpak export paths, and forwards it to the runuser xdg-open child like DISPLAY. The string builder is unit-tested; I could not exercise the elevated desktop path end to end, so a try on Ubuntu with snap Firefox would be welcome.

Fixes #1447

### DIFF
--- a/src/linux/platformquirks_linux.cpp
+++ b/src/linux/platformquirks_linux.cpp
@@ -198,6 +198,25 @@ namespace {
 
 namespace PlatformQuirks {
 
+// The XDG_DATA_DIRS a login shell would give this user: the spec default plus
+// the snap and flatpak export directories their profile scripts append.
+static void buildXdgDataDirs(const char* home, char* out, size_t outLen) {
+    std::snprintf(out, outLen,
+                  "/usr/local/share:/usr/share:/var/lib/snapd/desktop:"
+                  "/var/lib/flatpak/exports/share:%s/.local/share/flatpak/exports/share",
+                  home);
+}
+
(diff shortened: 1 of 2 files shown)


tdewey-rpi:
I took a variant of this change while coverage-grinding Imager - and credited you accordingly. Closing this PR as the other one has gone in.

PR #1732, opened 2026-09-14
fix(linux): reap io_uring completions while waiting for a ring buffer slot

### LINKED ISSUE #1731: [BUG]: Write stalled at writing 0% on 2.0.11.1 on debian trixie
### What happened?

I updated to the new 2.0.11.1

I use this to write custom .img to the internal SD reader in my laptop. Up to 2.0.10 this worked fine. I had 2.0.7 in my trash folder still so i could use that to finish work, same laptop, same boot session, same SD card.

On 2.0.11.1 it stalled with writing 0%.

On 2.0.10 and 2.0.7 Write and verified the image successfully.

I don't really know what to tell more. Logs from not working version and working version below. 

### Version

2.0.11.1 (Default)

### Host OS Version

Trixie

### Selected OS

Custom IMG

Raspberry Pi 3B, 3A+, 3B+, and Compute Module 3, 3+

### OS Customisation

### Relevant log output

```shell
2.0.11.1:

[FileOps] Device I/O limits: max_transfer=524288 bytes, suggested_queue_depth=128
Device authorization and open took 0 ms
Debug: Enabling direct I/O as per debug options
[FileOps] O_DIRECT enabled
Device max transfer: 524288 bytes
[FileOps] Async queue depth set to 256 (io_uring: yes)
Async I/O: configured with queue depth 256
SD card CID: "03534453443332478531aec6ea019c01"
SD card CSD: "400e00325b590000edc87f800a404001"
Zeroing out first and last MB of drive
WriteProgressWatchdog: Started monitoring
  First MB + flush took 42 ms
  Last MB + flush + sync took 46 ms
Done zero'ing out start and end of drive. Total MBR prep: 90 ms
PerformanceStats: Direct I/O state updated to enabled
File can be handled by libarchive as archive format
Started progress updates after successful drive opening
No compression filter detected (raw or uncompressed)
_writeFile: captured first block ( 524288 ) and advanced file offset via seek
forcePollAsyncCompletions: No completions available, still 256 pending
forcePollAsyncCompletions: No completions available, still 256 pending
forcePollAsyncCompletions: No completions available, still 256 pending
forcePollAsyncCompletions: No completions available, still 256 pending
forcePollAsyncCompletions: No completions available, still 256 pending
forcePollAsyncCompletions: No completions available, still 256 pending
forcePollAsyncCompletions: No completions available, still 256 pending
forcePollAsyncCompletions: No completions available, still 256 pending
forcePollAsyncCompletions: No completions available, still 256 pending
forcePollAsyncCompletions: No completions available, still 256 pending
forcePollAsyncCompletions: No completions available, still 256 pending
forcePollAsyncCompletions: No completions available, still 256 pending
[FileOps] High write latency detected (13142ms) - reducing queue depth to 128
[FileOps] Queue depth reduced for recovery: 256 -> 128 (pending: 256)
[FileOps] High write latency detected (13080ms) - reducing queue depth to 64
[FileOps] Queue depth reduced for recovery: 128 -> 64 (pending: 128)
[FileOps] High write latency detected (13054ms) - reducing queue depth to 32
[FileOps] Queue depth reduced for recovery: 64 -> 32 (pending: 64)
[FileOps] High write latency detected (13042ms) - reducing queue depth to 16
[FileOps] Queue depth reduced for recovery: 32 -> 16 (pending: 32)
[FileOps] High write latency detected (13035ms) - reducing queue depth to 8
[FileOps] Queue depth reduced for recovery: 16 -> 8 (pending: 16)
[FileOps] High write latency detected (13032ms) - reducing queue depth to 4
[FileOps] Queue depth reduced for recovery: 8 -> 4 (pending: 8)
[FileOps] High write latency detected (13030ms) - reducing queue depth to 2
[FileOps] Queue depth reduced for recovery: 4 -> 2 (pending: 4)


2.0.7:
Just works:
[FileOps] O_DIRECT enabled
Async I/O requested but not supported on this platform
SD card CID: "03534453443332478531aec6ea019c01"
SD card CSD: "400e00325b590000edc87f800a404001"
Zeroing out first and last MB of drive
WriteProgressWatchdog: Started monitoring
  First MB + flush took 41 ms
  Last MB + flush + sync took 75 ms
Done zero'ing out start and end of drive. Total MBR prep: 119 ms
PerformanceStats: Direct I/O state updated to enabled
File can be handled by libarchive as archive format
Started progress updates after successful drive opening
No compression filter detected (raw or uncompressed)
_writeFile: captured first block ( 8388608 ) and advanced file offset via seek
Write timing stats: writes= 224 syscall= 75151 ms preHashWait= 0 ms postHashWait= 0 ms sync= 0 ms syncCount= 0 avgSize= 8192 KB
Hash of uncompressed image: "f574174dba89914140c28808ff353f04962de71bfca1ef5185b6c6dd5ae0dc3a"
Write done in 78 seconds
Post-write verification using 6144 KB buffer for 1800 MB image
Verify hash: "f574174dba89914140c28808ff353f04962de71bfca1ef5185b6c6dd5ae0dc3a"
Verify done in 59.15 seconds
Checking customization: config= false cmdline= false firstrun= false cloudinit= false initFormat= "" isEmpty= true
Writing first block (which we skipped at first)
Write complete, switching to slow drive scanning
Drive scan mode changed to: Slow
unmountDisk: no mounts found for "/dev/mmcblk0"
Pipeline timing summary: decompress= 2771 ms (ring_wait= 0 ms)
Write timing stats: writes= 224 syscall= 75151 ms preHashWait= 0 ms postHashWait= 0 ms sync= 0 ms syncCount= 0 avgSize= 8192 KB
WriteProgressWatchdog: Stopped
PerformanceStats: Cycle ended, state: "succeeded" events: 18 samples: dl= 224 dec= 224 wr= 224 vfy= 299
```

yuefdev:
Fixes #1731.

## Problem

On 2.0.11.1 a write to an internal SD reader on Debian trixie stays at "Writing 0%". The same card, image and laptop work with 2.0.10 and 2.0.7. The reporter's log shows 256 io_uring writes queued, then `forcePollAsyncCompletions: No completions available, still 256 pending` once a second.

## Root cause

The write path can deadlock on Linux when the write ring buffer has no more slots than the async queue is deep:

1. With zero-copy async writes, each in-flight write holds a write ring buffer slot until its completion callback runs (`DownloadExtractThread::extractImageRun()` passes `releaseCallback` to `_writeFile()`).
2. The write ring buffer is capped at 256 slots (`maxWriteSlots` in `SystemMemoryManager::getCoordinatedRingBufferConfig()`). The async queue depth is also capped at 256, and reaches it when 8 GB or more of memory is available (`asyncQueueDepthFor()`). With 512 KB writes (`max_sectors_kb` of an mmc device), both come out at 256.
3. When slots == depth, the extract thread uses up every slot before `AsyncWriteSequential()` ever reaches its queue-full wait, which is the wait that reaps completions. Instead it waits in the slot loop in `extractImageRun()`, which only calls `_file->PollAsyncCompletions()`.
4. `LinuxFileOperations::PollAsyncCompletions()` has been a no-op since 499a842, because io_uring's completion queue must have a single consumer. So nothing reaps the completions, no callback runs, no slot is freed, and the write stays at 0% until the user cancels or the ring buffer stall timeout (90 s) fires.

## Why it first shows up in 2.0.11

The deadlock is also in 2.0.10. But earlier Linux AppImages were built without liburing, so they always wrote synchronously. The reporter's 2.0.7 log shows `Async I/O requested but not supported on this platform`. Since 4dfb06d / 720b900, AppImages are built in a bookworm chroot that has `liburing-dev`, so 2.0.11 is the first AppImage that uses io_uring.

## Fix

Keep the single-consumer rule, but let the thread that submits the writes reap from `PollAsyncCompletions()`:

- `AsyncWriteSequential()` records the calling thread in `cq_owner_thread_` (atomic `std::thread::id`) on every submission.
- `PollAsyncCompletions()` calls `ProcessCompletions(false)` only when it is called from that thread. Calls from other threads (the progress watchdog on the main thread, the download thread in `_pushQueue()`) are still no-ops, as 499a842 intended.

The extract thread's slot-wait loop now frees slots as writes complete. No other caller changes behaviour.

## Tests

Added `The writing thread reaps completions while it waits for a buffer` to `file_operations_test.cpp`. It fills a queue of depth 4, checks that a poll from another thread leaves the pending count alone (Linux only), then waits for the writes to drain by calling only `PollAsyncCompletions()` from the submitting thread, as the extract thread does. Without the fix the pending count never falls, and the test fails after its 10 s deadline.

I could not build or run this locally: I only have a Windows machine without WSL or Docker available. The change has not been compiled or run yet. It would be good if CI, or someone on Linux with liburing >= 2.2, could run:

```sh
cmake --build build --target file_operations_test && ./build/test/file_operations_test "[file-ops]"
```

## Workaround for affected users

Open the debug options with Ctrl+Alt+S and turn off "Enable Async I/O", or set the queue depth below 256.

## Not addressed here

In the reporter's log, the latency-based depth reduction in `ProcessCompletions()` drops the depth from 256 to 2 in one batch. When many old completions are reaped together, each one passes the `currentPending <= async_queue_depth_` check again as the pending count falls. This does not stall the write, but it may leave throughput lower than intended for the rest of the write. It could be a separate change.


tdewey-rpi:
Fix looks correct, some additional points that need to be addressed before PR merged:

- Before merging: make the new test Linux-only with #ifdef __linux__ or a SKIP on other platforms. On macOS the writes finish without polling, so REQUIRE(inFlight > 0) will probably fail.
- In the test: declare buffers and returned before ops, so a failed REQUIRE can't cause a use-after-free.
- In Close(): reset cq_owner_thread_ to std::thread::id{}.
- Simplify: reduce the PollAsyncCompletions() check to just the owner-thread comparison.
- Correct: the comment on _pushQueue() and the PR description's claim that no other caller changes behaviour.
- Follow-up: add a DownloadExtractThread test with write slots equal to queue depth, covering the actual #1731 path.

PR #1734, opened 2026-09-14
fix(i18n): load the regional translation for a regional locale

### LINKED ISSUE #1572: [BUG]: Brazilian Portuguese translation not working
### What happened?

Brazilian Portuguese translation not working

When opening version 2.0.7, the following appears in the pt-pt translation

Image

### Version

2.0.7 (Default)

Windows

### Host OS Version

Windows 11 25h2

### Selected OS

Windows

### OS Customisation

### Relevant log output

```shell

```

yuefdev:
Fixes #1572.

## Problem

Brazilian Portuguese users get the European Portuguese translation, and they can't choose Brazilian Portuguese from the language menu either. There are two separate causes.

**At startup**, `main()` loads the translation with `QTranslator::load(QLocale(), "rpi-imager", "_", ":/i18n")`. That looks for files whose locale part uses underscores, such as `rpi-imager_pt_BR.qm`. The regional translations are named with hyphens (`rpi-imager_pt-BR.qm`, `rpi-imager_zh-TW.qm`, `rpi-imager_en-US.qm`), so Qt never finds them and falls back to the bare language. Replaying that call against the compiled translations with Qt 6.11.2:

| locale | file loaded |
|---|---|
| pt_BR | `rpi-imager_pt.qm` (European Portuguese) |
| zh_TW | `rpi-imager_zh.qm` (Simplified Chinese) |
| en_US | `rpi-imager_en.qm` |

**In the language menu**, `ImageWriter` keys `_translations` by each translation's native language name. `QLocale("pt")` and `QLocale("pt-BR")` both give "português". `rpi-imager_pt-BR.qm` sorts first and `pt` then overwrites it, so the menu has 26 entries for 27 translations. zh-TW is not affected, because Qt names it "繁體中文" and zh "简体中文".

## Fix

- **Picking a translation.** `ImageWriter::translationForLocale()` matches the locale's `uiLanguages()` against the translation codes. Those are already hyphenated BCP 47 tags. Whole tags are tried first, then each tag cut back one subtag at a time.
  - `main()` now calls `ImageWriter::setLanguageForLocale(QLocale())`. It installs the match and records it as the current language.
  - The constructor preselects the current language with the same function. Before, it compared only the part of `QLocale::name()` before the underscore.
- **Naming in the menu.** When a regional translation has the same name as another translation, the region is added to its name: "português (Brasil)". "português" keeps its name, so a saved choice of it still applies. No other name changes.
- **Falling back.** A regional translation is installed over its language's translation, pt-BR over pt and zh-TW over zh. Anything it hasn't translated yet then comes out in that language instead of English.
  - This matters for zh-TW. It has 19 finished translations out of 745, while zh has 539. Without zh underneath, Taiwan would now get an almost entirely English UI.
  - `replaceTranslator()` removes the whole set when the language changes. It is still there for `--qm`.

## Changes in behaviour worth knowing about

- **en_US now gets `en-US` on top of `en`.** Today that changes one string: "Capital city:" becomes "National capital:". That is what the en-US translation says, but it never loaded before.
- **Download telemetry.** `_currentLangcode` is passed to download telemetry. It now carries the regional code, for example `pt-BR`, when the locale matches one. Before, a regional code only got there when someone picked the language by hand.

## Tests

New cases in `src/test/image_writer_test.cpp`, all tagged `[imagewriter][locale]`:

- `Every translation is in the language menu`: one menu entry per `.qm` in `:/i18n`.
- `Brazilian Portuguese is named apart from Portuguese`
- `A regional locale gets its regional translation`: `translationForLocale()` for pt_BR, zh_TW, en_US, pt_PT, zh_CN, en_GB, de_AT and fi_FI.
- `Starting in Brazil selects Brazilian Portuguese`
- `A regional translation falls back on its language`: after choosing 繁體中文, a string only zh translates still comes out translated. After switching to English it comes out in English again.

How I ran them: `image_writer_test` does not build on Windows because it includes `<pwd.h>`, and I don't have a Linux machine to hand. So I built on Windows 11 with MinGW 13.1 and Qt 6.10.2 (Release). Two local changes were needed for that build, and neither is part of this PR:

- `main` does not currently compile on Windows (`src/rpiboot/file_server.cpp:533`). I applied the fix from #1733 on top.
- I kept `wlanapi_delayed.lib` out of `rpi_imager_testable`'s sources, because the test tree does not configure on Windows with it.

- **The new cases and the six existing `[imagewriter][locale]` cases.** I copied them unchanged into a small Catch2 runner, with the same `main()` as `image_writer_test`, linked against `rpi_imager_testable`.
  - With this change: 11 test cases, 29 assertions, all pass.
  - Against `main`'s `imagewriter.cpp`, `imagewriter.h` and `main.cpp`, I could only run the two new cases that compile there. Both fail: `langs.size() == files.size()` gives `26 == 27`, and "português (Brasil)" is not in the menu. The existing cases pass.
- **`qml_ui_test` on `tst_language_step.qml`**: 12 passed, 0 failed.
- **The full `qml_ui_test` suite**: 1075 passed, 17 failed and 21 skipped, with exactly the same 17 failures as `main` on the same machine. They are Windows-specific and in code this change doesn't touch: file url handling, `WizardNavigation` and `SecureBootStep`.
- **`rpi-imager` itself builds.**

It has not been run on Linux or macOS.


tdewey-rpi:
- British and Portuguese Linux users get the wrong regional translation. With the common setting LANGUAGE=en_GB:en, Qt 6.11.1 lists en-US ahead of plain en. The new matcher ([imagewriter.cpp:4525](vscode-webview://0v8v3r1e0smkb4cfhe6i4sa7nhgpdet2kau814lp2dqo439j0q5t/src/imagewriter.cpp#L4525)) takes the first whole match, so UK users start in American English. LANGUAGE=pt_PT:pt gives Brazilian Portuguese in the same way. I reproduced both with a small program against that Qt. The new tests don't catch it because QLocale("en_GB") built by hand doesn't add those entries.
- Embedded mode loses its "English" default. Under LANG=C, [imagewriter.cpp:370](vscode-webview://0v8v3r1e0smkb4cfhe6i4sa7nhgpdet2kau814lp2dqo439j0q5t/src/imagewriter.cpp#L370) now overwrites it with an empty string, which main's old code didn't do.

tdewey-rpi:
Merged via my PR that packaged some fixups. Thanks for the contribution, @yuefdev

PR #1735, opened 2026-09-14
fix(drivelist): name internal eMMC as eMMC, not an SD card reader

### LINKED ISSUE #1658: [BUG]: Internal eMMC appearing as "Internal SD card reader"
### What happened?

Incredibly misleading. Can cause someone to overwrite their system drive. Speaking from experience. Something like "Internal eMMC Device" would be significantly more accurate and would prevent things like this from happening.

### Version

2.0.10 (Default)

### Host OS Version

not relevant

### Selected OS

not relevant

Raspberry Pi Zero 2 W

### OS Customisation

### Relevant log output

```shell

```

yuefdev:
Fixes #1658: an unnamed eMMC drive was labelled as an internal SD card reader on Linux.

## Change

Identify eMMC from its `mmcblkNbootM` hardware boot partitions and label it `Internal eMMC storage`. Unnamed SD readers now get their fallback label at any `mmcblkN` index, including an SD card at `mmcblk1` alongside eMMC at `mmcblk0`.

Existing vendor/model labels, device filtering and removability rules are preserved. Boot partitions remain excluded from the list. This adds one untranslated source string; `.ts` files are unchanged.

## Verification

All 95 assertions in `Linux lsblk parsing` pass, including both device orderings and the existing parser cases. Restoring the `mmcblk0` restriction makes the new reversed-layout test fail.

The Linux parser and its unchanged test case were compiled in a standalone Windows harness using MSVC 2022, Qt 6.10.2 and Catch2. This verifies JSON parsing, not Linux hardware enumeration. No physical eMMC or RasPiKey testing was performed.

AI-assisted implementation and PR wording.


src/drivelist/drivelist_test.cpp
```diff
@@ -498,6 +498,157 @@ TEST_CASE("Linux lsblk parsing", "[drivelist][linux][unit]")
...
+            }]
+        })";
+
+        auto devices = parseLinuxBlockDevices(json, false);
+
+        REQUIRE(devices.size() == 2);
+        CHECK(devices[0].device == "/dev/mmcblk0");
+        CHECK(devices[0].description == "Internal SD card reader");
+        CHECK(devices[1].device == "/dev/mmcblk1");
+        CHECK(devices[1].description == "Internal eMMC storage");
```
lurch, line 649: Depending on how the hardware is wired up, could these also occur the other way around, i.e. "Internal eMMC storage" at `/dev/mmcblk0` and "Internal SD card reader" at `/dev/mmcblk1` ? :thinking: 

And thinking of exotic edge-cases, I wonder if there are any systems that have **two** "Internal eMMC storage" devices?

yuefdev: Yes, that happens. The index follows the device tree `mmc` aliases, or probe order where there are none, so some boards put the eMMC at `mmcblk0` and the SD slot at `mmcblk1`.

With this PR that layout comes out as:

- `mmcblk0` (has `mmcblk0boot0`/`boot1`): "Internal eMMC storage"
- `mmcblk1` (SD reader, no vendor/model): no description

The second line is unchanged from `main`, because the "Internal SD card reader" label has only ever applied to `mmcblk0`. What changes is that on `main` the eMMC at `mmcblk0` is the device labelled "Internal SD card reader".

Now that eMMC is detected on its own, the `mmcblk0` restriction could go, and any `mmcblkN` without boot partitions could be labelled "Internal SD card reader". I left that out to keep the change narrow. I'm happy to add it here, with a test section for this layout, if you'd like.

On two eMMC devices: `findEmmcDevices()` collects every device that has boot partitions, so each one would get "Internal eMMC storage". They'd still be told apart by device path and size in the list. I don't know of a board that has two, but I can add a test section for that case too.

lurch: > The second line is unchanged from `main`, because the "Internal SD card reader" label has only ever applied to `mmcblk0`.

I _suspect_ that may have been an oversight? (Because we've probably never tested RPi Imager on a platform that has both "Internal eMMC storage" **and** "Internal SD card reader"!)

> Now that eMMC is detected on its own, the `mmcblk0` restriction could go, and any `mmcblkN` without boot partitions could be labelled "Internal SD card reader". I left that out to keep the change narrow. I'm happy to add it here, with a test section for this layout, if you'd like.

I think that would make sense :slightly_smiling_face: 

> On two eMMC devices: `findEmmcDevices()` collects every device that has boot partitions, so each one would get "Internal eMMC storage". They'd still be told apart by device path and size in the list. I don't know of a board that has two, but I can add a test section for that case too.

Probably best to wait and see what @tdewey-rpi says.

yuefdev: Thanks, that makes sense. I'll remove the mmcblk0 restriction and add a test with eMMC at mmcblk0 and an SD card at mmcblk1. I'll wait for @tdewey-rpi 's input before adding the two-eMMC test case.

tdewey-rpi: > On two eMMC devices: findEmmcDevices() collects every device that has boot partitions, so each one would get "Internal eMMC storage". They'd still be told apart by device path and size in the list. I don't know of a board that has two, but I can add a test section for that case too.

This is somewhat-tricky. A device like the RasPiKey (from UUGear) would be pretty interesting to test - though how you'd do that on a Raspberry Pi with a single SD interface while _also_ having a second eMMC available would be effectively impossible.

I think we could probably live with both being labelled internal, at least until we encounter real-world feedback that is causing a problem.

Don't let perfect be the enemy of good, etc.

tdewey-rpi:
I'm not generally opposed to AI-assisted coding, but I would _strongly_ request that you push the LLM to be terse in it's output - each of these PRs comes with a wall of text that isn't helpful in rapid analysis.

I'd recommend two directives to your LLM:

1) Limit sections to 100 words maximum
2) Use British English (though that's just my preference!)

In practice, (1) helps significantly with code reviews - and with your token budgets too!

src/drivelist/drivelist_test.cpp
lurch: > A device like the RasPiKey (from UUGear)

Interesting. I wonder if that'd be seen as an "eMMC device" by Linux (i.e. whether the `mmcblk0boot0` and `mmcblk0boot1` devices would be exposed) or if it'd be seen as just a "plain micro SD card".

yuefdev:
(Posted by mistake, please ignore.)

src/drivelist/drivelist_test.cpp
yuefdev: Removed the mmcblk0 restriction and added the reversed-layout test in 3afa055. All 95 parser assertions pass in the Windows Qt harness. Agreed on keeping both eMMC devices labelled internal for now. I cannot confirm how RasPiKey exposes its boot partitions without hardware output.

yuefdev:
Thanks, I have shortened the descriptions here and in #1737, and saved both preferences for future replies.

src/drivelist/drivelist_linux.cpp
```diff
@@ -188,9 +188,10 @@ std::optional<DeviceDescriptor> parseBlockDevice(const QJsonObject& bdev, bool e
     // Internal MMC devices have no vendor or model to go by. eMMC is soldered
     // to the board, often as the system drive, so it must not be passed off
     // as a card reader: that invites overwriting it.
+    static const QRegularExpression mmcDevice(QStringLiteral("^/dev/mmcblk\\d+$"));
     if (emmcDevices.contains(name) && descParts.isEmpty()) {
         descParts.append(QObject::tr("Internal eMMC storage"));
-    } else if (name == "/dev/mmcblk0" && descParts.isEmpty()) {
+    } else if (mmcDevice.match(name).hasMatch() && descParts.isEmpty()) {
         descParts.append(QObject::tr("Internal SD card reader"));
     }
```
lurch, line 207: I guess one of these is a subset of the other, so _perhaps_ it might make sense to change this to:
```cpp
    static const QRegularExpression mmcDevice(QStringLiteral("^/dev/mmcblk\\d+$"));
    if (mmcDevice.match(name).hasMatch() && descParts.isEmpty()) {
        if (emmcDevices.contains(name)) {
            descParts.append(QObject::tr("Internal eMMC storage"));
        } else {
            descParts.append(QObject::tr("Internal SD card reader"));
        }
    }
```
:thinking:

src/drivelist/drivelist_linux.cpp
```diff
@@ -60,10 +62,34 @@ void walkStorageChildren(DeviceDescriptor& device, QStringList& labels, const QJ
...
+/**
+ * @brief Find the MMC devices that are eMMC rather than SD cards
+ *
+ * eMMC has hardware boot partitions, which the kernel lists as disks of their
+ * own (mmcblk0boot0, mmcblk0boot1) next to the device they belong to. An SD
+ * card never has them.
+ */
+QSet<QString> findEmmcDevices(const QJsonArray& blockDevices)
+{
+    static const QRegularExpression bootPartition(QStringLiteral("^(/dev/mmcblk\\d+)boot\\d+$"));
```
lurch: Possibly outside the scope of the PR, but I guess this regex could also be used on line 110? :wink:

src/drivelist/drivelist_linux.cpp
yuefdev: Done in 41bf815: the eMMC check is now nested inside the `mmcblkN` match, as suggested.

src/drivelist/drivelist_linux.cpp
yuefdev: Done in 41bf815: the regex now lives in `emmcBootPartition()`, which is used both to find eMMC devices and to skip the boot partitions (replacing the `contains("boot")` check).

PR #1736, opened 2026-09-14
fix(qml): turn a picked file's url into the path it names

yuefdev:
## Problem

The in-app file pickers (`ImFileDialog`) hand back a `file://` url. Five callers turn it into a path by cutting `file://` off the front:

- `SshKeyManager.qml`: "Browse" for an SSH public key file.
- `AppOptionsDialog.qml`: the secure-boot signing key.
- `SecureBootCustomizationStep.qml`: the same key, from the wizard.
- `DebugOptionsDialog.qml`: a custom fastboot gadget.
- `main.qml`: the performance-data save dialog.

That doesn't give a path.

- **On every platform**, the url's escaping stays in. A `#` in the name comes out as `%23` and a `%` as `%25`, so `id#1.pub` is looked for as `id%231.pub`.
- **On Windows**, `file:///C:/Users/x/key.pub` becomes `/C:/Users/x/key.pub`, which does not open. Qt resolves it to `C:/C:/Users/x/key.pub`.

These pickers aren't used everywhere:

- They are the fallback when there is no native dialog: embedded mode, a Linux session without a desktop portal, or `--qml-file-dialogs`.
- They are the only picker for the fastboot gadget in the debug options.

So the effect is narrower than the code suggests. Picked through one of these pickers:

- A file whose name needs escaping is not found. The SSH key picker has worked this way since 2.0.2 and the signing-key pickers since 2.0.0.
- On Windows, no file is found at all. Since the gadget picker has no native alternative, a custom gadget can't be chosen on Windows.

On Windows the QML suite shows the problem. `test_a_key_file_is_read_and_its_keys_added`, `test_choosing_a_file_twice_does_not_double_the_keys` and `test_choosing_a_gadget_stores_a_path_rather_than_a_url` fail with:

```
Failed to open file: "/C:/Users/.../authorized_keys" Error: "The filename, directory name, or volume label syntax is incorrect."
```

## Fix

`ImageWriter::localPathFromUrl()` returns `url.toString(QUrl::PreferLocalFile)`: the local path for a `file:` url, and anything else as it came. All five callers now use it.

`test_saving_performance_data_writes_the_file_that_was_chosen` built its url as `"file://" + TestFiles.localPath(name)`. On Windows that makes the drive letter the host. It now uses the url `TestFiles.write()` returns, the same form the dialog hands back.

## Tests

- `image_writer_test.cpp`: `The url a file picker hands back becomes the path it names`. It round-trips real files named `id_ed25519.pub`, `id#1.pub`, `100% key.pub` and `clé.pub`, and checks that each one reads back. It also checks that an empty url gives an empty path and that a non-file url is left alone.
- `tst_ssh_key_manager.qml`: a key file named `id#1.pub` is read.
- `tst_app_options.qml` and `tst_secure_boot_step.qml`: a signing key url with `%23` is recorded as the `#` path.

The `#` cases fail against `main` on any platform. The existing Windows failures listed above are the Windows side.

How I ran them: Windows 11, MinGW 13.1, Qt 6.10.2, Release. `main` needs #1733 to build on Windows, and the test tree needs `wlanapi_delayed.lib` kept out of `rpi_imager_testable`'s sources to configure. Neither local change is part of this PR. `image_writer_test` itself does not build on Windows because of `<pwd.h>`, so I ran its two file cases, unchanged, in a small runner linked against `rpi_imager_testable`.

| | `main` | with this change |
|---|---|---|
| `tst_ssh_key_manager.qml` | 27 passed, 3 failed | 30 passed |
| `tst_app_options.qml` | 18 passed, 1 failed | 19 passed |
| `tst_secure_boot_step.qml` | 8 passed, 3 failed | 9 passed, 2 failed * |
| `tst_debug_options.qml` | 35 passed, 1 failed | 36 passed |
| `tst_main_window.qml` | 34 passed, 1 failed | 35 passed |
| `Reading a file the user picked` + the new C++ case | (new case does not compile) | 2 cases, 30 assertions pass |
| full `qml_ui_test` | 1075 passed, 17 failed, 21 skipped | 1082 passed, 13 failed, 21 skipped |

Across the full suite, the four Windows failures above now pass, and so do the three new cases. None of the 13 still failing involves a picked file:

- 11 are local-repository cases (`OSDelegate`, `WizardNavigation`), fixed separately in #1737.
- 2 are the `SecureBootStep` cases in the footnote.

\* The two left are `test_turning_secure_boot_on_survives_the_session` and `test_turning_it_back_off_removes_what_was_recorded`. They fail on Windows before and after this change. The registry hands the setting back as the string `"true"`, and `compare()` wants the boolean. The step itself accepts both.

It has not been run on Linux or macOS.


tdewey-rpi:
I fed this to Claude, who fixed up the merge conflicts from my coverage-grinding sessions:

- The in-app file dialog turns a typed path into a url by gluing "file://" onto the front without escaping. In perf#1.json, the # is then read as the start of a url fragment.
- QUrl::toString(PreferLocalFile) only converts a url to a path when it has no fragment, so the new helper hands back the whole file:///…#1.json string instead of a path.
- A throwaway test confirmed it: the old approach of cutting off file:// gave the right path, the new one doesn't.
- The performance-data save dialog is hit hardest, because the name there is always typed.
- The fix belongs in the dialog's _toFileUrl (ImFileDialog.qml:87), which should escape the path, or in the helper (imagewriter.cpp:522), which could return the local path whenever the url is a local file.

Issue #1609, opened 2026-04-30
[BUG]: Error using local_json manifest: "Unable to download OS list"


Cristian-Radiantis:
### What happened?

I generated a local manifest using:

py create_local_json.py --online --capabilities usb_otg --device-capabilities usb_otg

This correctly creates:
os_list_local.rpi-imager-manifest

However, when trying to use this file in Raspberry Pi Imager (v2.0.7, Windows), Imager shows:

"Unable to download OS list. You can still use a local image file."

No OS list is displayed, only "Erase" and "Use custom image".

It seems Imager is not able to load the generated manifest and falls back to offline mode.

Is there any additional step required to properly use the local manifest?

### Version

2.0.7

### What host operating system were you using?

Windows

### Host OS Version

Windows 11 Pro 25 H2

### Selected OS

Raspberry Pi Os Lite (64 bit)

### Which Raspberry Pi Device are you using?

Raspberry Pi 5, 500, and Compute Modules 5

### What kind of storage device are you using?

microSD Card in an internal reader

### OS Customisation

- [x] Yes, I was using OS Customisation when the bug occurred.

### Relevant log output

```shell

```

tdewey-rpi:
How did you try to use the local manifest?

Cristian-Radiantis:
Hi! I tried both ways, directly opening the file which ends up opening the rpi imager and by directly opening from the settings in the rpi imager.

tdewey-rpi:
I've just tried this with 2.0.9, and cannot reproduce the bug. Are you able to try with the pre-release 2.0.9? https://github.com/raspberrypi/rpi-imager/releases/tag/v2.0.9

lurch:
@Cristian-Radiantis Do you actually have any OSes listed in your `os_list_local.rpi-imager-manifest` file?

PR #1737, opened 2026-09-14
fix(fetch): distinguish URL normalisation from redirects

yuefdev:
Fix curl URL normalisation being mistaken for a redirect, which breaks subsequent local repository refreshes on Windows. Possibly related to #1609; that report has not been reproduced directly.

## Change

Use `CURLINFO_REDIRECT_COUNT` before reporting the effective URL. Curl can normalise a `file:///C:/...` URL without following a redirect; saving that normalised URL corrupts later fetches. Real HTTP redirects still update the repository URL.

## Verification

Previously tested on Windows 11 with MinGW 13.1 and Qt 6.10.2:
- Fetcher tests: 17 passed, including local-file normalisation and HTTP redirects.
- `tst_os_delegate.qml`: 14 passed.
- Full QML suite: 1095 passed, 6 failed, 12 skipped. Remaining failures concern file URLs and Secure Boot settings.

The fetcher harness used a CA-bundle stub and a local Python server. Windows builds needed #1733 and test-linking workarounds. Linux and macOS were not tested.

AI-assisted implementation and PR wording.


tdewey-rpi:
Could I please ask you to reword this PR - the first line isn't super descriptive - for one _what_?

yuefdev:
Reworded the title and shortened the description. It now makes clear that curl URL normalisation was being mistaken for a redirect.

tdewey-rpi:
- Keep both checks, but replace the duplicated Windows comments with one-line reasons in curlfetcher.cpp and imagewriter.cpp.
- Remove the unused #include <cstring>.
- Add an http test: a /./ URL on the local server keeps its requested URL. That's the case this PR now changes in production.
- Turn the /./ test's REQUIRE precondition into a SKIP, and note that the file:// test only catches the regression on Windows.
- Rebase onto main and reword the description: main's 67c882fe already fixes the Windows symptom, so this fixes the cause and stops http(s) URLs being rewritten.

